Issue #23, opened 2019-04-22
React Native support


AndrewMorsillo:
It looks like there are just a couple little missing pieces to work with react native. I'm not sure exactly how to integrate them properly into the codebase so I will put the snippets here. We just need a NetworkStatus implementation and to pass AsyncStorage for the storage engine when creating the client. IMO these should just be defaults when react native is the environment.

Detecting react-native
```
if (typeof navigator != 'undefined' && navigator.product == 'ReactNative') {
  // I'm in react-native, use the ReactNativeNetworkStatus and AsyncStorage
}
```

React native network status implementation
```
import {  NetInfo } from "react-native"
import { NetworkStatus, NetworkStatusChangeCallback } from "apollo-offline-client"

class ReactNativeNetworkStatus implements NetworkStatus {
  public onStatusChangeListener(callback: NetworkStatusChangeCallback): void {
    const listener = (connected: boolean) => {
      callback.onStatusChange({online: connected})
    };
    NetInfo.isConnected.addEventListener('connectionChange', listener)
  }

  public async isOffline(): Promise<boolean> {
    const isConnected = await NetInfo.isConnected.fetch()
    return !isConnected
  }
}
```

Config for react native
```
import { AsyncStorage } from "react-native"
import ReactNativeNetworkStatus from "./react-native-network-status.ts"

let config = {
  httpUrl: "http://localhost:4000/graphql",
  wsUrl: "ws://localhost:4000/graphql",
  storage: AsyncStorage,
  networkStatus: new ReactNativeNetworkStatus()
} 
```

edit: oops it looks like AsyncStorage might not be just drop in. Types don't line up but I think it will work regardless.

xtagon:
Not everyone is using React let alone React Native. I use Ember.js with Apollo, for example, and plan to use Offix with Ember.js as well.

Is there a simple way to support this feature for React Native without introducing a dependency on React Native libraries?

AndrewMorsillo:
That is why I opened the issue here. I'm not sure how to have easier react native setup without adding a dependency.

xtagon:
I would vote for adding it to the docs. (Note: I'm not the maintainer, just an outside opinion).

If the configuration gets more complicated than that, another idea is to create a separate integration library to handle some of the setup that is specific to React Native. For example, this is what ember-apollo-client does for Ember + Apollo integration, and you could probably do something similar for React Native + Offix integration.

wtrocki:
[...]

PR #24, opened 2019-04-22
React native package

wtrocki:
Created to collaborate on the dedicated package for react native.

#23


wtrocki:
Closing as there were no updates. We plan react native support in near future

PR #26, opened 2019-04-22
Rename packages to conform with the new project name (offix)

wtrocki:
We should rename project to new packages name as soon as possible to avoid confusing community and clear up architecture.

Target package names:

- offix-client 
- offix-offline 
- offix-conflicts
- offix-client-react-native
- offix-client-cordova

- offix-server (node.js)


wtrocki:
FYI @StephenCoady

Issue #6, opened 2019-04-15
Delta sync


ntziolis:
A generalized way to handle changes in data on the server while client was offline without refetching all data after reconnect as well as transparently execute initial data loading without requiring user to have all data locally already for app to function.

I completely understand that this is not a client only feature but this library seemed to be the best place to start a discussion on how we would go about establishing a standardized way to achieve this capability (server to client data replication) when leveraging a graphql backend

**Describe the solution you'd like**
- Client side API that allows to specify which data should be delta synced from the server to the client and how in a non boilerplate manner
- Server side requirements to support delta sync
- Establish overridable conventions

**Describe alternatives you've considered**
- AWSAppSync
  - Its AWS only
  - No support for apollo >2.5 local state management and local resolvers
- Custom built on graphql
  - Well ... it would be custom building a central component of what a lot of folks need
- custom build on top of puchdb
  -  requires couchdb compatible db on server vs graphql endpoint

wtrocki:
@ntziolis actually this was our target from the day one and going to be covering this requirement by:

- Implementing a new cache layer on top of Apollo GraphQL InMemory storage.
Working on Open Sourcing that part in https://github.com/graphql-heroes/ApolloQLCached

- Integrating QLCached with ApolloOfflineClient.

It is really glad to see this request coming from the community as it validates that this use case can be useful for people. General target will be to support diffs in OfflineClient. Going to create Roadmap issue soon so we can put some timing on when and how things will be delivered. 

@ntziolis Do you have time to collaborate on requirements/aproach for this?

wtrocki:
Forgot to mention that we are not only client side and js. 
Our goal is to provide a comprehensive layer for both Client and Server side. 
Currently, we provide server-side package for conflicts and planning to do more in that space soon.

ntziolis:
This is just awesome. I'd love to assist in this. My background is in building replication engines so lmk how I can participate. I think true offline capabilities that do NOT force single cloud vendor lock-in or custom solutions are the last major piece in the GraphQL all the way puzzle.

Since I now seem to be talkig to the like minded I wanne run something by you I have been thinking about for a long time now:

Assuming we have a way todo delta sync in a transparent way. One also needs a way to query data (incl filtering etc) on the client side. While this could be implemented manually in custom client side resolvers with apollos new client side state handling it would lead to a lot of handcrafted boilerplate code. 
Instead I'd love to build something like Prisma/joinmonster etc but for the client side.

[...]

PR #32, opened 2019-04-29
Decouple client into separate class

wtrocki:
## Motivation

Make offline api in separate object while still attaching helpers to Apollo client. Give developers ability to override the way that client is created:
for example enable custom cache etc. get access to storage or register listeners. 
This change is backward compatible with the rest of the API.
f81e2e3


wtrocki:
👍 @StephenCoady for quick review.  
Ping @ntziolis - Going to drop another change soon that will provide the groundwork for #6.

PR #34, opened 2019-04-30
Enable pluggable listeners for offline changes

wtrocki:
## Motivation

Allow registering multiple listeners for detecting offline queue changes like an item added etc. Additionally, we will be enabling registering listeners on the fly as opposed to the one single startup listener


wtrocki:
I'm going to hold of with documentation updates and going to create general chapter in #35

PR #44, opened 2019-05-27
feat: support multiple queries, export functions

StephenCoady:
### Description

This PR provides the ability to update multiple queries in the cache when a mutation is run.

It also exports valuable functions the client may want to use without using the full cache helper function.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


wtrocki:
Great work!

PR #46, opened 2019-06-07
feat: split helpers into its own package

StephenCoady:
### Description

To make the mutation and subscription helpers more usable they should be in their own package. I have taken them from the client and also removed some of the logic used by the helpers to helper functions within this package.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


examples/conflicts/package.json
```diff
@@ -16,7 +16,8 @@
     "axios": "^0.18.0",
     "express": "^4.16.4",
     "graphql": "0.13.2",
-    "graphql-tools": "^4.0.3"
+    "graphql-tools": "^4.0.3",
+    "offix-helpers": "^0.1.0"
```
wtrocki: How about `offix-cache` or something that will be more descriptive

wtrocki:
@StephenCoady I will play with that on master and make release

packages/offix-server/package.json
```diff
@@ -31,5 +31,8 @@
   "publishConfig": {
     "access": "public"
   },
-  "gitHead": "c52429fc496cfdbaaf06fd55a509f14b12c7ce1c"
+  "gitHead": "c52429fc496cfdbaaf06fd55a509f14b12c7ce1c",
+  "dependencies": {
+    "offix-cache": "^0.1.0"
```
wtrocki: Do we need that here?

StephenCoady: nope, i was messing with lerna it must have added it. will remove

PR #48, opened 2019-06-12
Readme and delete bug

StephenCoady:
### Description

We need a readme for this project so I have added that.

Also there was a bug where a delete did not remove from the cache. If a server only returns the id of the object being deleted then we need to cater for that. 

Interface improvements to decouple optimistic response from mutation helpers.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


packages/offix-cache/README.md
```diff
@@ -1,5 +1,140 @@
...
+  mutation: ADD_TASK,
+  variables: { some_key: "some_value"},
+  operationType: "add",
+  returnType: "Task",
+  idField: "id"
+})
+
+apolloClient.mutate({
+  mutation: ADD_TASK,
+  optimisticResponse
```
wtrocki: This is actually es6 feature so maybe we can 
```suggestion
  optimisticResponse: optimisticResponse
```

packages/offix-client/test/createOptimisticResponseTest.ts
```diff
@@ -1,37 +0,0 @@
-import { createOptimisticResponse, MutationHelperOptions, CacheOperation  } from "offix-cache";
-import { expect, should } from "chai";
-import { CREATE_ITEM, GET_ITEMS } from "./mock/mutations";
-
-describe("Optimistic Response tests", () => {
```
wtrocki: ?? :>

StephenCoady: because these tests are in the cache package now, not the client - I just forgot to move this file last time.

PR #54, opened 2019-06-14
Conflict refinement

wtrocki:
Changing API with breaking changes to handle conflicts on clients only. 
Interfaces were simplified to remove sever side conflicts.
We also can now directly throw error to users because there is no actual choice between client and server. This makes API much cleaner. 
Server-side still dictates the API (pluggability)

We are moving from api
```
    // 1. Read data from data source
    // 2. Check for conflicts
    if (conflictHandler.hasConflict(serverData,clientData)) {
      // 3. Resolve conflict (client or server) and return error to client
      return await conflictHandler.resolveOnClient(serverData, clientData).response
    }
    // 4. Call next state to update
    greeting = conflictHandler.nextState(clientData)
    // 5. Save object to data source
``` 
to 
```
    // 1. Read data from data source
    // 2. Check for conflicts
    conflictHandler.checkForConflicts(serverData,clientData)
    // 3. Save object to data source

```

## Progress 

- [x] Modifications in API
- [x] Integration Tests


wtrocki:
@StephenCoady @darahayes ready for quick review

wtrocki:
published as `offix-conflicts-server`

packages/offix-conflicts-server/package.json
```diff
@@ -1,12 +1,12 @@
 {
-  "name": "offix-server",
-  "version": "0.6.0",
+  "name": "offix-conflicts-server",
```
darahayes: perhaps `offix-server-conflicts?` and then we will be able to release other packages in the future with the pattern `offix-server-<feature>` if necessary?

packages/offix-server-conflicts/src/states/VersionedObjectState.ts
```diff
@@ -0,0 +1,53 @@
...
+ *   version: String
+ * }
+ */
+export class VersionedObjectState implements ObjectState {
+
+  public checkForConflict(serverState: ObjectStateData, clientState: ObjectStateData) {
+    if (serverState.version && clientState.version) {
+      if (serverState.version !== clientState.version) {
+        const filteredServerState: any = {};
+        for (const field in clientState) {
```
darahayes: Minor nit but I think this method of looping of an object's own properties is kinda strange. I think that this is more readable

```
for (const key in Object.keys(clientState)) {
 // filteredServerState[field] = serverState[field];
}
```

No big deal either way though :)

wtrocki: Good advice. we need the same on client

wtrocki:
Ready for review. Can be verified with showcase

wtrocki:
Done and ready for review. I have this linked to showcase

docs/sections/ref-conflict-server.md
```diff
@@ -41,7 +41,7 @@ type Greeting {
     msg: String
     version: Int
   }
-```
+```n b
```
darahayes: Was this accidental?

packages/offix-server-conflicts/src/states/VersionedObjectState.ts
```diff
@@ -0,0 +1,52 @@
...
+export class VersionedObjectState implements ObjectState {
+
+  public checkForConflict(serverState: ObjectStateData, clientState: ObjectStateData) {
+    if (serverState.version && clientState.version) {
+      if (serverState.version !== clientState.version) {
+        const filteredServerState: any = {};
+        for (const key of Object.keys(clientState)) {
+          filteredServerState[key] = serverState[key];
+        }
+        throw new ObjectConflictError({
```
darahayes: The more I think about this, the stronger I feel it is not the best approach. The function is called `checkForConflict` and if one is found, it instantly throws. I feel like this is a highly opinionated approach and is probably a strong anti-pattern.

I think it will come back to bite us. The fact we are throwing an error like this is based off our current understanding of Apollo Server error handling, which has changed in the past and may change again. In a way, this is creating a loose dependency on Apollo Server internals and the user does not even realise it. Also, just in general it's really weird that the user would explicitly ask 'Is there a conflict?' and we say 'Yes, here is the info' via throwing an error.

As an end user, I would not expect an error to be thrown by this function. I think we should instead return the error object and tell the user to either return/throw it themselves. My understanding is you don't even need to throw because Apollo Server checks the return value and will do the same error handling there too.

This would mean the user would write some code like this.

```js
const conflict = conflictHandler.checkForConflict(serverData,clientData)

if (conflict) {
  // as an end user/developer I can perform any other tasks I need as part of conflicts
  // logging
  // update metrics/monitoring
  // send email to boss :)

  throw conflict // or just return it
}

// rest of resolver
```

I think this approach would be much more beneficial.

* The code written by the end user is much more explicit and readable. The intention is clear.
* The user can have additional steps for dealing with conflicts (logging, metrics, send email to the boss) without having to write unnecessary `try/catch` blocks everywhere in their resolvers.
* We do not make any assumptions in our internal code about how errors are handled by the GraphQL environment. apollo-server vs graphql-js vs graphql-express etc.

@wtrocki @StephenCoady what do you think?

wtrocki: Brilliant idea. Going to apply this now

wtrocki:
@darahayes What will be your preference here:

A) Keep package in voyager pointing to this?
B) Remove package from voyager?


Ready for final review

wtrocki:
Verified, but blocked by documentation changes in #68

wtrocki:
documentation landed. i think we might merge it and refine docs in separate task.

wtrocki:
Ready for review

docs/ref-conflict-server.md
```diff
@@ -51,67 +51,25 @@ Every conflict can be handled using a set of predefined steps
...
-The`ObjectState` interface is a complete conflict resolution implementation that provides a set of rules to detect and handle conflict. Interface will allow developers to handle conflict on the client or the server. `nextSate` method is a way for interface to modify existing object before is being saved to the database.
+The`ObjectState` interface is a complete conflict resolution implementation that provides a set of rules to detect and handle conflict. Interface will allow developers to handle conflict on the client. Client side application will need to match the server side implementation 
 For example when using `lastModified` field as a way to detect conflicts:
 
 ```typescript
- public nextState(currentObjectState: ObjectStateData) {
-    currentObjectState.lastModified = new Date()
-    return currentObjectState
+  public checkForConflict(serverState: ObjectStateData, clientState: ObjectStateData) {
+    if(serverState.lastModified === clientState.lastModified){
```
darahayes: I might be missing something here but this looks funny to me. This is an example of custom conflict resolution but would it not make sense to be throwing the `ObjectConflictError` when those values are not equal? i.e. `if(serverState.lastModified !== clientState.lastModified)` ??

packages/offix-server-conflicts/src/api/ObjectConflictError.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Conflict error that is being returned when server
+ * Error specific to Voyager framework
+ */
+export class ObjectConflictError extends Error {
+  public conflictInfo: ConflictData;
+
+  constructor(data: ConflictData) {
+    super("VoyagerConflict");
```
darahayes: Still want the name `Voyager` in here?

docs/ref-conflict-server.md
```diff
@@ -51,67 +51,25 @@ Every conflict can be handled using a set of predefined steps
 ```javascript
     // 1. Read data from data source
     // 2. Check for conflicts
-    if (conflictHandler.hasConflict(serverData,clientData)) {
-      // 3. Resolve conflict (client or server) and return error to client
-      return await conflictHandler.resolveOnClient(serverData, clientData).response
-    }
-    // 4. Call next state to update
-    greeting = conflictHandler.nextState(clientData)
+    conflictHandler.checkForConflicts(serverData,clientData)
```
darahayes: Should we update this with something like this?

```js
const conflict = conflictHandler.checkForConflict(serverData, clientData)
if(conflict){
  throw conflict;
}
```

packages/offix-server-conflicts/src/api/ObjectConflictError.ts
wtrocki: @Stephen WDYT about DataConflict?

darahayes: I do like the idea of having something like `VoyagerConflict` or `OffixConflict` or `OffixDataConflict` because it is a specific non-generic name for a specific thing that is slightly outside of the realm of regular GraphQL. As an end user that may not know things inside out, this at least makes it very clear in their application logs that they are dealing with something from offix. Helps lead them in the right direction quicker. WDYT?

StephenCoady: Yeah +1 to something specific. `OffixConflict` is good for me.

wtrocki:
@StephenCoady Done - we can alter client as well

PR #67, opened 2019-06-24
Add integration test for offlineMutation

b1zzu:
### Description

I found some unexpected behaviors when updating the cache using the `offlineMutation()` method that could be bugs so I have created a few integration tests to reproduce this behavior. 

My only doubt is if I'm using wrongly the library or there is actually one or more bugs.

This PR is meant to show the errors, and not to be merged. The fact that the tests are failing is wanted for now.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-cache/src/cache/mutations.ts
+++ b/packages/offix-cache/src/cache/mutations.ts
@@ -109,29 +109,32 @@ export const getUpdateFunction = (options: CacheUpdateOptions): MutationUpdaterF
   switch (operationType) {
     case CacheOperation.ADD:
       updateFunction = (cache, { data }) => {
-        try {
-          if (data) {
-            let queryResult = cache.readQuery({ query, variables }) as any;
-            const operationData = data[operation];
-            const result = queryResult[queryField];
-            if (result && operationData) {
-              if (!result.find((item: any) => {
-                return item[idField] === operationData[idField];
-              })) {
-                result.push(operationData);
-              }
-            } else {
-              queryResult = [result];
-            }
-            cache.writeQuery({
-              query,
-              variables,
-              data: queryResult
-            });
+        let queryResult;
+        if (data) {
+          const operationData = data[operation];
+          try {
+            queryResult = cache.readQuery({ query, variables }) as any;
+          } catch (e) {
+            queryResult = {};
           }
-        } catch (e) {
-          console.info(e);
+
+          const result = queryResult[queryField];
+          if (result && operationData) {
+            if (!result.find((item: any) => {
+              return item[idField] === operationData[idField];
+            })) {
+              result.push(operationData);
+            }
+          } else {
+            queryResult[queryField] = [result];
+          }
+          cache.writeQuery({
+            query,
+            variables,
+            data: queryResult
+          });
         }
+
       };
       break;
     case CacheOperation.DELETE:
--- a/packages/offix-cache/src/cache/mutations.ts
+++ b/packages/offix-cache/src/cache/mutations.ts
@@ -1,7 +1,7 @@
 import { MutationOptions, MutationUpdaterFn, OperationVariables } from "apollo-client";
 import { CacheOperation } from "./CacheOperation";
 import { createOptimisticResponse } from "../optimisticResponse";
-import { Query } from "./CacheUpdates";
+import { CacheUpdatesQuery } from "./CacheUpdates";
 import { getOperationFieldName, deconstructQuery } from "../utils/helperFunctions";
 import { isArray } from "util";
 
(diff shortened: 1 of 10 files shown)


wtrocki:
This is something we definietly should take look. I will add that to my queue.

ping @StephenCoady

wtrocki:
@b1zzu Awesome work! I'm going to take look into now and push fixes here :D

wtrocki:
#Closed by mistake

## `using only updateQuery fails`

The current mechanism we have is using `mutationCacheUpdates` method for every mutation when offline. If this is not passed there will be no updates provided:

It is a very important bug that can be resolved by:


1) caching update method in the link and supplying that in mutate method after.
This approach is 100% correct for the live app scenario but is going to fail after a restart when the original method is gone.

2) Instead of requiring this parameter everytime we can simply save all input from updateQuery field and use those to build update method in internal mutate that is executed (assuming that user did not provide custom value)

3) ???

@sebastienblanc need your input on this.


## `without cache initialization fails`

IMHO we should check if the query exists before and attempt to create it.

wtrocki:
@StephenCoady FYI (any preference) ?

wtrocki:
Actually to fix first problem using aproach 2 we will need to have the following parameters all helper parameters saved in offlineQueue. We have already saving typeName in offline queue so we can also save name of the id field and we could be sorted.

b1zzu:
@wtrocki I'm pro the 2. solution but we have to make clear to the user that it will not work if he provide a custom update method

wtrocki:
Additional obvious problem is that we do not support single item responses (expects array all the time so 

`getProfile: Profile`is not going to work

wtrocki:
@b1zzu Awesome! Leave that with me :)

wtrocki:
I have made couple tests with:

- Using single object 
- Using fragment
- Using variables
- Using relationships

Screenshot 2019-06-25 at 17 31 00

I think that it will be hard to get coverage for all the cases here. Going to try to find the best approach.

wtrocki:
@b1zzu I needed to move out from this temporarily because we need to get conflict work first. The conflict has some improvements that can be used to resolve this problem. I do not want to merge that now as we need some actions from this.

b1zzu:
> Do you think we should merge that?

Sorry, @wtrocki I don't understand what are you talking about, because if you are referring to the tests it doesn't make sense for me to merge them because they are failing, or would you like to merge only the passing tests?

wtrocki:
@b1zzu Missunderstanding. This PR is passing build on CircleCI

b1zzu:
@wtrocki I've forgotten a `.only` inside the tests, so it was running only one test, but there are few tests that are passing should I commit them in a sperate PR?

wtrocki:
ah.. ignore me then. I will push fix here

wtrocki:
TL;DR:

- When starting with an empty cache we cannot create item and update 
- We do not know if we are dealing with Array or Single return object (there is no way to get it from query or variables. We can check the type of the cache element, but if query is used for the first time then we missing that information.
- For cache helpers to work with offline we need to always pass them globally, which kinda defeats the purpose and makes api much harder.

wtrocki:
This PR contains a lot of different issues we need to address. 

For 0.7.0 we can do:

- Documentation that global cache updates should be provided.
- Documentation for cache helpers support only arrays
- Fix problem with

b1zzu:
I would also add:
* Document how to initialize the cache the first time if the device is offline

wtrocki:
@StephenCoady THere are still couple concerns here that are not addressed but I'm hoping to get that merged (working on docs now) and build another PR that will refactor the workflow.

wtrocki:
Working on docs now.

wtrocki:
Ready for review

packages/offix-cache/src/cache/mutations.ts
```diff
@@ -53,113 +96,112 @@ export const createMutationOptions = <T = {
...
+            variables,
+            data: queryResult
+          });
         }
+
       };
       break;
     case CacheOperation.DELETE:
       updateFunction = (cache, { data }) => {
-        try {
```
StephenCoady: any reason you removed the try from here? any cache action should be wrapped in a try/catch block surely.

wtrocki: ah.. right. I removed it because the only case I found for failure was on reading data, but then delete can also fail on read. This is bug

wtrocki:
@StephenCoady I have played with this on showcase (separate PR was created for showcase problem) and managed to get most of the integration tests done.

@b1zzu - you are true star here! THX for hard work on tests.

PR #68, opened 2019-06-24
fix: Move documentation to docsify

wtrocki:
Use docsify as a documentation framework. 
No content changes.


darahayes:
Just going to drop a suggestion for an alternative. https://docusaurus.io

Some benefits IMO

* Statically Generated Pages - Docsify is a single page so it's hard to search the web for specific topics our docs might cover
* React based - Any customisation, styling etc will be done in an environment/framework we're all somewhat comfortable with. Bonus: we can take advantage of that and possibly add demos/interactive elements to the docs site. Kinda like how ionic does it.
* Well maintained. Docsify is an amazing project but the maintenance and updates are fairly slow right now.

wtrocki:
@darahayes Going to refactor contect to docusaurus.io then. 
We might apply it then to all our open source projects based on spike.
Ping @ankitjena

Issue #74, opened 2019-06-29
Bug: id.startsWith is not a function


Hasan-git:
![image](https://user-images.githubusercontent.com/8820596/60386212-eb031680-9a9a-11e9-9b2c-d3890f489c67.png)

wtrocki:
This is happening when the model is missing `id` field or id field is not a string
@Hasan-git I will quickly patch it as this looks like a blocker for me.

I think we have 2 issues here. The first one is that we always assume `id` is a string and this is a critical issue that I will patch today or tomorrow.

The second one will be the fact that we ignore `MutationHelperOptions.idFIeld` provided by the user and use the hardcoded id in `IDProcessor` - Going to create a separate issue for that bug.

Can you provide me a value of the id field that is triggering this issue to make sure that my fix is going to address it?

To fix this bug we need:
- Make additional check for id and transform it to string
- Document the MutationHelperOptions.idFIeld for customization needs

Hasan-git:
> Can you provide me a value of the id field that is triggering this issue to make sure that my fix is going to address it?

Indeed, the whole object was null. if its does matter that I was doing a mutation that should return an GraphqlError but I didn't ,  maybe if I was returning an GraphqlError this bug will not be catched

wtrocki:
@Hasan-git This makes sense now. Going to patch this and do new patch release.

wtrocki:
Done. Going to release 0.6.1

PR #77, opened 2019-07-01
fix: Refactor id processor to always use toString

wtrocki:
fix for https://github.com/aerogear/offix/issues/74

Waiting for more info to write integration tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "coverage": "nyc report --reporter=text-lcov | coveralls",
     "cleanInstall": "lerna exec npm install --ignore-scripts",
     "bootstrap": "lerna bootstrap --no-ci",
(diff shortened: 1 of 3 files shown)


darahayes:
@wtrocki looks like your editor did some auto formatting. Was that intentional?

wtrocki:
@darahayes - Yes. Some of the files we have are not formatted properly to avoid problems with backporting changes I have formatted it (sorry for readability)

darahayes:
@wtrocki @StephenCoady it would be nice to have some unit tests for the `execute` function here. From what I can see it's deterministic on the inputs provided so it should be easy to unit test. What would you think about creating an issue and getting @MStokluska to have a look?

wtrocki:
@StephenCoady Quick review.

PR #79, opened 2019-07-02
feat:  field level conflicts

StephenCoady:
### Description

Field Level Conflict Resolution

This PR takes the following approach:

We can record changes on each individual client as a base state. This state is saved to the operation's context and persisted in the offline queue. If a conflict happens then we can use that information to calculate diffs. This reduces the amount of data that needs to be saved and gives simplicity to end users.

Pros
Works almost out of the box on server
Handling logic is very simple to understand and implement for developers

Architecture:

Screenshot 2019-07-02 12 09 17

##### Checklist

- [ ] Documentation changes
- [x] Tests


packages/offix-cache/src/cache/mutations.ts
```diff
@@ -5,9 +5,15 @@ import { Query } from "./CacheUpdates";
 import { getOperationFieldName, deconstructQuery } from "../utils/helperFunctions";
 import { isArray } from "util";
 
+export interface Context {
+  conflictBase?: any;
+  returnType?: string;
```
wtrocki: I do not understand this.
We have explicit fields added bellow (idField and itType).
Is the fact that we attaching those to context internal implementation specific.
Missing comment with motivation for this interface

packages/offix-client/src/config/DataSyncConfig.ts
```diff
@@ -82,7 +76,7 @@ export interface DataSyncConfig {
    * Interface that defines how object state is progressed
    * This interface needs to match state provider supplied on server.
    */
-  conflictStateProvider?: ObjectState;
+  conflictProvider?: ObjectState;
```
wtrocki: We need to add notes to release notes for those

packages/offix-client/src/conflicts/ConflictLink.ts
```diff
@@ -0,0 +1,178 @@
...
+export class LocalConflictError extends Error {
+  constructor(private base: any, private variables: any) {
+    super();
+  }
+}
+
+/**
+ * Represents conflict information that was returned from server
+ */
+export interface ConflictInfo {
```
darahayes: Potential Overlap with the interface defined here? https://github.com/aerogear/offix/pull/54/files#diff-b8f9c4f82651357a985e9cfd8207cd4eR9

Should we deduplicate these things into a commmon package at some point such that we guarantee any interfaces share between client and server match up? I've been burned by this before. Defining the same interface in two different places. Had to make changes later and ended up with runtime errors that were not caught at compile time. Not saying let's do it right now but it's food for thought!

packages/offix-client/src/conflicts/ConflictLink.ts
```diff
@@ -0,0 +1,178 @@
...
+    this.strategy = this.config.conflictStrategy;
+    this.listener = this.config.conflictListener;
+  }
+
+  public request(
+    operation: Operation,
+    forward: NextLink
+  ): Observable<FetchResult> | null {
+    if (isMutation(operation)) {
+      if (this.stater.currentState(operation.variables) !== undefined) {
```
darahayes: Minor suggestion: Up until this point the code was really easy to understand. 😅 It's probably just me but even some minor comments adding some context would be nice in here. Especially if the ability to do handle conflicts is one of our core features. Having a small bit of context in the code would be nice for anyone else interested in potentially contributing.

packages/offix-client/src/conflicts/ConflictLink.ts
```diff
@@ -0,0 +1,178 @@
...
+  /**
+   * The conflict resolution strategy your client should use. By default it takes client version.
+   */
+  conflictStrategy?: ConflictResolutionStrategy;
+}
+
+/**
+ * Conflict handling link implementation that provides ability to determine
+ */
+export class ConflictLink extends ApolloLink {
```
wtrocki: It will be cool to get more documentation. This is the main Link class.

packages/offix-client/src/links/LinksBuilder.ts
```diff
@@ -60,26 +74,28 @@ export const createOfflineLink = async (config: DataSyncConfig, store: OfflineSt
...
   const mutationOfflineLink = ApolloLink.split((op: Operation) => {
     return isMutation(op) && !isOnlineOnly(op);
   }, offlineLink);
-  const retryLink = ApolloLink.split(OfflineMutationsHandler.isMarkedOffline, new RetryLink(config.retryOptions));
-  const localFilterLink = new LocalDirectiveFilterLink();
-  const links: ApolloLink[] = [mutationOfflineLink, retryLink, localFilterLink];
+  const baseLink = new BaseLink(config.conflictProvider as ObjectState, cache);
+  const links: ApolloLink[] = [baseLink, mutationOfflineLink];
+  links.push(conflictLink);
+  // const retryLink = ApolloLink.split(OfflineMutationsHandler.isMarkedOffline, new RetryLink(config.retryOptions));
```
wtrocki: Uncomment that :D

packages/offix-client/src/conflicts/ConflictLink.ts
```diff
@@ -0,0 +1,178 @@
...
+    operation: Operation,
+    forward: NextLink
+  ): Observable<FetchResult> | null {
+    if (isMutation(operation)) {
+      if (this.stater.currentState(operation.variables) !== undefined) {
+        return new Observable(observer => {
+          let retriedResult: any;
+          let sub: any;
+          let retriedSub: any;
+          sub = forward(operation).subscribe({
```
wtrocki: This code is ok, but it is quite hard to read and unit test.
If we extract observable logic into a separate class (retried observable or something). 
It is fine to leave it but mixing observables and business logic is why Apollo Client source code is really hard to read and understand.

packages/offix-client/src/conflicts/handler/ConflictHandlerOptions.ts
```diff
@@ -0,0 +1,42 @@
+import { ConflictResolutionData } from "../strategies/ConflictResolutionData";
+import { ConflictResolutionStrategy, ConflictListener } from "..";
+import { ObjectState } from "../state/ObjectState";
+
+/**
+ *
```
wtrocki: Missing comment

packages/offix-client/src/conflicts/strategies/ConflictResolutionStrategy.ts
```diff
@@ -0,0 +1,20 @@
+import { ConflictResolutionData } from "./ConflictResolutionData";
+
+/**
+ * Interface for strategy that can be used to resolve conflict
+ */
+export interface ConflictResolutionStrategy {
+  /**
+   *
+   * Strategy resolution method
```
wtrocki: Improve this comment

packages/offix-client/src/links/LinksBuilder.ts
wtrocki: Uncomment this code

packages/offix-client/src/conflicts/handler/ConflictHandler.ts
```diff
@@ -0,0 +1,82 @@
...
+  /**
+   * Executes supplied strategy
+   *
+   * @param operationName - name of operation that triggered that conflict
+   */
+  public executeStrategy() {
+    const resolvedData = this.options.strategy.resolve(this.options.base,
+                                                      this.serverDiff, this.clientDiff, this.options.operationName);
+    this.options.objectState.assignServerState(resolvedData, this.options.server);
+    if (this.options.listener) {
```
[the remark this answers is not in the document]
wtrocki: -1. We do not run this in Node.js environment :)

packages/offix-client/src/conflicts/ConflictLink.ts
StephenCoady: I actually figured out how to use onError for our specific case instead of just dropping the code in here like this. Also makes it way easier to test. Pushing soon.

docs/ref-conflict-server.md
```diff
@@ -34,7 +34,7 @@ type Greeting {
     msg: String
     version: Int
   }
-```
+```n b
```
wtrocki: Remove this

PR #96, opened 2019-07-08
feat: offix offline package

wtrocki:
New package structure:
Screenshot 2019-07-08 at 23 42 49

Done:

- Reorganize imports into api
- Extract a couple of classes into separate files.
- Get package published manually and get permissions sorted.
- Remove circular dependencies
- Remove invalid ideas and experiments


packages/offix-offline/package.json
```diff
@@ -0,0 +1,63 @@
+{
+  "name": "offix-offline",
```
wtrocki: @StephenCoady @darahayes Do you like to change package.json name? 
Let me know beforehand :D

StephenCoady: Its fine with me anyway.

PR #103, opened 2019-07-10
General improvements

wtrocki:
I went thru the code and removed bad smells and other problems that were related with using legacy api or bad naming. There are no controversial changes.

- UseClient was documented in docs but we exposed clientWins - clientWins is different approach than UseClient so this required rename.
- Renamed DataSyncConfig 
- Removed mobile core dependency that brings up 300kb into a bundle (sick!) 
- Renamed folders for circleci builds
- Removed old legacy references to our fork of cache persist.

### DIFF
--- a/packages/offix-client/src/OfflineClient.ts
+++ b/packages/offix-client/src/OfflineClient.ts
@@ -1,6 +1,6 @@
 import { ApolloClient } from "apollo-client";
-import { DataSyncConfig } from "./config";
-import { SyncConfig } from "./config/SyncConfig";
+import { OffixClientConfig } from "./config";
+import { OffixDefaultConfig } from "./config/OffixDefaultConfig";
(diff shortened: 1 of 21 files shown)


wtrocki:
@StephenCoady WDYT?

packages/offix-client/src/OfflineClient.ts
```diff
@@ -1,6 +1,6 @@
 import { ApolloClient } from "apollo-client";
-import { DataSyncConfig } from "./config";
-import { SyncConfig } from "./config/SyncConfig";
+import { OffixClientConfig } from "./config";
+import { OffixDefaultConfig } from "./config/OffixDefaultConfig";
```
StephenCoady: Finally! Now I won't get confused between the interface and the implementation any more 👍

wtrocki: Took we a while to figure out the name but yeah! 😿 So happy that you like it!

PR #104, opened 2019-07-10
Update Apollo GraphQL packages

wtrocki:
Working on fix for #104


wtrocki:
@StephenCoady I think that generally having merged types with Apollo will cause a lot of problems over time. My first thought is to separate them and operate on plain Apollo options and Separate cache helper options. 

Cache helper options are strongly typed with generics that are not used in MutationHelperOptions extensions. WDYT?

StephenCoady:
So you mean just rely fully on plain mutate and supply the `update` function separately using a new endpoint? So we need to split and use `optimisticResponse` and `update` individually then. If you think that will decrease problems in the future with being tied to Apollo's API then I am all for it.

wtrocki:
I think we still decorate their original object but we do not supply additional parameters with it. The fact that we supply all additional parameters means that we need to conform to the generic interface - which causes some issues.

StephenCoady:
OK, sounds fine to me. let me throw that into a pr now

StephenCoady:
@wtrocki see here: https://github.com/aerogear/offix/pull/106

I broke the merged type - do you think an intersection is enough for what you mean? I know top level is still the same but this way they can move on independently.

wtrocki:
Closing

PR #105, opened 2019-07-10
fix: Refactor package.json depedencies

wtrocki:
Reorganize package structure so it will have only the required dependencies in place. Do not reimport packages as this will cause a mess for clients


packages/offix-offline/src/offline/LocalDirectiveFilterLink.ts
```diff
@@ -3,7 +3,7 @@ import { hasDirectives, removeDirectivesFromDocument } from "apollo-utilities";
 import { localDirectivesArray, MUTATION_QUEUE_LOGGER } from "../utils/Constants";
 import * as debug from "debug";
 
-export const logger = debug.default(MUTATION_QUEUE_LOGGER);
+const logger = debug.default(MUTATION_QUEUE_LOGGER);
```
wtrocki: This was actually exporting debug type - Top level apps will not work without @types/debug :)

wtrocki:
@StephenCoady This can potentially cause conflicts for your PR but it will not affect it that much.

PR #106, opened 2019-07-10
fix: break hard dependency on mutation options interface

StephenCoady:
Enable type safety for offlineMutation by conforming to Apollo mutate interface.


packages/offix-cache/src/cache/mutations.ts
```diff
@@ -20,7 +20,7 @@ export interface MutationHelperOptions extends MutationOptions {
  * Provides useful helpers for cache updates, optimistic responses, and context
  * @param options see `MutationHelperOptions`
  */
-export const createMutationOptions = (options: MutationHelperOptions): MutationOptions => {
+export const createMutationOptions = (options: MutationOptions & MutationHelperOptions): MutationOptions => {
```
wtrocki: ah.. that is even nicer. This will decorate MutationHelperOptions and still work with <T> depedency.

wtrocki:
@StephenCoady Can you also bump versions of the packages to see if this is going to compile with latest Apollo 2.6?

packages/offix-client/src/ApolloOfflineClient.ts
```diff
@@ -28,6 +28,6 @@ export interface ApolloOfflineClient extends ApolloClient<NormalizedCacheObject>
    * Allows the client to perform an offline mutation
    * @param options the mutation helper options used to build the offline mutation
    */
-  offlineMutation<T>(options: MutationHelperOptions): Promise<FetchResult<T>>;
+  offlineMutation<T>(options: MutationOptions & MutationHelperOptions): Promise<FetchResult<T>>;
```
StephenCoady: @wtrocki we should probably reuse the merged type

packages/offix-client/src/OfflineClient.ts
```diff
@@ -96,7 +98,7 @@ export class OfflineClient implements ListenerProvider {
    * this offline friendly function to handle the optimistic UI and cache updates.
    * @param options the MutationHelperOptions to create the mutation
    */
-  public offlineMutation<T>(options: MutationHelperOptions): Promise<FetchResult<T>> {
+  public offlineMutation<T>(options: MutationOptions & MutationHelperOptions): Promise<FetchResult<T>> {
```
StephenCoady: @wtrocki and here :P

PR #110, opened 2019-07-11
mutationCacheUpdates integration test fix

jhellar:
### Description

The part of integration test for mutationCacheUpdates should check that cache was updated after client restarts. The tests passes now, but should actually do the check with `cache-first` query.


wtrocki:
Awesome change 💯 ! We totally missed this.

Issue #111, opened 2019-07-11
getUpdateFunction does not have a default opType


StephenCoady:
## Bug Report

Currently `getUpdateFunction` here: https://github.com/aerogear/offix/blob/master/packages/offix-cache/src/cache/mutations.ts#L77 does not have a default value for `opType`. This is bad and can be breaking.

* **Affected Package(s)**:
offix-cache
* **Package version(s)**:
1.6.1

StephenCoady:
fixed in https://github.com/aerogear/offix/pull/117

PR #112, opened 2019-07-11
fix: add default opType to mutation options

StephenCoady:
### Description

This is a fix for https://github.com/aerogear/offix/issues/111

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-cache/src/cache/mutations.ts
+++ b/packages/offix-cache/src/cache/mutations.ts
@@ -74,8 +74,12 @@ export const createMutationOptions = <T = {
 export const getUpdateFunction = (
   operation: string,
   idField: string,
-  opType: CacheOperation,
+  opType?: CacheOperation,
   updateQuery?: Query): MutationUpdaterFn => {
+
+  if (!opType) {
+    opType = CacheOperation.ADD;
+  }
 
   if (!updateQuery) {
     return () => {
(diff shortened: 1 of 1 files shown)


wtrocki:
Ah.. but this requires undefined to be passed. Kinda defeats the purpose.

StephenCoady:
ah. theres no way to do it then without breaking the api really.

wtrocki:
yeah, rule of thumb is to add things at the end with defaults or use objects if possible. Objects are cool for typescript but hard for js so they should be used sporadically.

StephenCoady:
@wtrocki if we want the updateQuery to be optional and to provide a default for the opType we don't have an option but to use objects really. wdyt?

wtrocki:
Generally rule of thumb from clean code is to never have more than 3 inputs to function. If we do more then this function becomes unmanagable. It is even worse if the function becomes an public api. Objects are probably for the best

StephenCoady:
@wtrocki I have made the changes. I know you are hesitant for a breaking change but it is here if you want it. I will leave it up to you whether we merge this now or after.

wtrocki:
>  know you are hesitant for a breaking change 

Absolutely not. We cannot avoid breaking here

StephenCoady:
Closing this as these changes are included in https://github.com/aerogear/offix/pull/117

Issue #116, opened 2019-07-15
Default UseClient conflict strategy can pick outdated data


StephenCoady:
## Bug Report

Currently, conflict processor is creating an issue where it updates issues in the queue meaning conflicts (that should be triggered) are not.

Also, the strategy interface should be extended to allow the passing of additional necessary data. 

The default implementation is also incorrect and has the potential to cause data loss as server data is overwritten with base data. To combat this the base should not be used in the resolution as it is not needed. Instead we can apply the `clientDiff` data to the server data returned with the conflict. This ensures only up to date data is passed with the next mutation in the queue.

Currently working on this.

* **Affected Package(s)**: offix-offline

StephenCoady:
Fixed in https://github.com/aerogear/offix/pull/117

PR #117, opened 2019-07-15
fix: default strategy no longer uses out dated data

StephenCoady:
breaking: interfaces updated to accept a more complete data. now accepts an object.
breaking: getUpdateFunction updated to make certain vars optional. now accepts an object.

### Description

addresses https://github.com/aerogear/offix/issues/116 and https://github.com/aerogear/offix/issues/111

there are breaking changes in this PR.

updates to integration tests, fyi @jhellar

also @wtrocki @darahayes I have added yarn to the scripts - do we still want to do that? I am +100. I haven't added it as a dev dependency, let me know if you want me to.

EDIT: just saw on a separate PR you want to just go with npm but alias it locally. Shout here if you want me to change yarn -> npm on the files here.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "test": "nyc lerna run test",
     "coverage": "nyc report --reporter=text-lcov | coveralls",
-    "cleanInstall": "lerna exec npm install --ignore-scripts",
+    "cleanInstall": "lerna exec yarn install --ignore-scripts",
```
wtrocki: ❤️ . 

@darahayes Any voice against it?

wtrocki:
Yarn link was so convinient. Tried and tested. With this change every subsequent offline edit ends with conflict. I'm not sure we should be documenting this as this is going to get addressed soon anyway.

packages/offix-offline/src/conflicts/strategies/ConflictResolutionStrategy.ts
```diff
@@ -1,21 +1,14 @@
...
-   * @param base - base data before any client changes were applied
-   * @param server - the latest server data returned to the client
-   * @param client - the data the client tried to send to the server
-   * @param operation - [optional] the name of the operation you wish to perform the strategy on.
    */
-  resolve: (base: ConflictResolutionData,
-            server: ConflictResolutionData,
-            client: ConflictResolutionData,
-            operation?: string) => ConflictResolutionData;
+  resolve: (data: ConflictDataSet) => ConflictResolutionData;
```
wtrocki: Set is usually considered and array like object. How about just `ConflictData`?

wtrocki: No strong opinion :P

package.json
darahayes: Coming to this late but go for it :)

PR #122, opened 2019-07-18
fix: stop using hardcoded id field for operations.

StephenCoady:
this commit fixes a problem where we assume id is the unique
field for an object. to fix this i have used the idField which
we already have available. if it is not provided in context then
we revert to the default of "id"

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/offix-cache/src/cache/mutations.ts
+++ b/packages/offix-cache/src/cache/mutations.ts
@@ -122,7 +122,7 @@ export const createMutationOptions = <T = {
     options.update = update;
   }
 
-  options.context = { ...context, returnType };
+  options.context = { ...context, returnType, idField };
   return options;
 };
 
--- a/packages/offix-client/integration_test/test/offline.test.js
+++ b/packages/offix-client/integration_test/test/offline.test.js
@@ -459,6 +459,48 @@ describe('Offline mutations', function () {
     });
   });
 
+  describe('ensure offline params are added to offline object ', function () {
+
+    let task;
+
+    beforeEach('prepare data', async function () {
+      client = await newClient({ networkStatus, storage: store, mutationsQueueName });
+      networkStatus.setOnline(true);
+
+      const response = await client.mutate({
+        mutation: ADD_TASK,
+        variables: newTask,
+        optimisticResponse: createOptimisticResponse(optimisticOptions),
+      });
+
+      task = response.data.createTask;
+
+      networkStatus.setOnline(false);
+    });
+
+    it('should succeed', async function () {
+      const variables = { title: 'nomerge1', description: 'nomerge', id: task.id, version: task.version };
+      try {
+        await client.offlineMutation({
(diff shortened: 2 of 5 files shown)


packages/offix-cache/src/cache/mutations.ts
```diff
@@ -75,7 +75,7 @@ export const createMutationOptions = <T = {
     }
   };
   options.update = update;
-  options.context = { ...context, returnType };
+  options.context = { ...context, returnType, idField };
```
wtrocki: Would this work only when we use the helper?

StephenCoady: yes, but `offlineMutation` in the client uses this function. we already say in our docs that if you want to use conflicts you need to provide `returnType` to `offlineMutation` or directly in the context so it would be the same here.

Do you have any other ideas on how to do this?

wtrocki: It is on client.cache.config.dataId from object?

https://github.com/apollographql/apollo-client/blob/master/packages/apollo-cache-inmemory/src/inMemoryCache.ts#L204

StephenCoady: @wtrocki it looks like this function uses __typename from _result_. From my understanding, we won't have access to anything that can give us the id field at this point. The first time we will have access is by using the optimistic response as a result, which uses the `idField` variable anyway.

Correct me if I am wrong

wtrocki: Ok.. so we have helpers that do not support many use cases for the moment. 
And then we have id mapping and base feature that relies on the value from helpers. We looking for full outer join for these cases. 

For base logic, we actually have both typename and id (and can use this function under the hood. For id mapping we have access to result that has everything to calculate the id (as this stuff will land in cache seconds after mapping is done).

wtrocki: our helpers accept `id` which is typical field and we have cache accessible it witch is `__typename+id` by default but it can be whatever you want - including new Date()

wtrocki: Thinking of it telling people - use `id` is probably too much

wtrocki:
Ok.. So following the approach that we require people to use offlineMutate. This is ok. Id uses id by default and it is going to swap it to something else. 
I have played with this and in fact, cacheUpdates here are not required - it will just create an optimistic response but when the custom optimistic response is provided it will still work. 

I was hoping we can use already existing method for fetching the id based on the entire object (optimistic response for base) and actual response. 

There is an case that we do not support here - when someone do not use `id` field do not want to use optimisticReponse and his id is not fiel but it is generated by function. This should be simple fix we can follow up with a better approach later.
 
TL;DR - played with it and fix is working for most of the cases we going to support. 
It will be cool to add more docs.

Issue #120, opened 2019-07-17
Network state should be checked before hitting client mutate


wtrocki:
## Definition of the problem

Currently, when performing offline mutate operation we going to apply optimistic response updates twice. This approach also requires users to pass the global cache update function, which is not convenient. 

## Solution

We can return error to the user and check network state directly in the helper making. Offline link then will be able to enqueue the changes as always but this time we going to know that this operation was marked as offline. 
This will help us to introduce server state aware offline cases where developers can specify some function that will quality changes to be enqueued to offline queue after network request.

wtrocki:
Ok.. After quick investigation of this I think that we can move even farther from client.mutate and make network state checkbeforehand. This will mean that there will be extra offlineFlag to context that will be triggering offline link to store mutations. I could not find any side impacts of this change and going to create quick prototype now

wtrocki:
This is already addressed in https://github.com/aerogear/offix/pull/170

wtrocki:
ping @darahayes

PR #125, opened 2019-07-19
fix: Move network check before calling client mutate

wtrocki:
Creating for early feedback. 
Addresses #120 

I'm still not sure about this step but it has tons of advantages:

- entire flow is very simple and easy to document now
- there is no additional corner cases and issues with other links - conflicts for example
- Improved performance
- There is no need to pass global cache updates (and when persisting optimistic responses we can remove that feature completely later)

Dissadvantages:

- Good bye client.mutate for good 
- Tell me more? 

Going to work on addressing the tests and couple concerns and making global refactor but if you do not like the idea let's kill it.


wtrocki:
The more advanced state now with documentation updates. Still missing notice about breaking changes and rename of the method.

wtrocki:
Checking integration tests :)

packages/offix-client/src/OfflineClient.ts
```diff
@@ -100,21 +101,27 @@ export class OfflineClient implements ListenerProvider {
...
    */
-  public offlineMutation<T = any, TVariables = OperationVariables>(
+  public offlineMutate<T = any, TVariables = OperationVariables>(
     options: MutationHelperOptions<T, TVariables>): Promise<FetchResult<T>> {
     if (!this.apolloClient) {
       throw new Error("Apollo offline client not initialised before mutation called.");
+    }
+
+    const mutationOptions = createMutationOptions<T, TVariables>(options);
+    if (this.config.networkStatus.isOffline) {
```
StephenCoady: this should be 
await this.config.networkStatus.isOffline()

packages/offix-client/src/OfflineClient.ts
```diff
@@ -100,21 +101,27 @@ export class OfflineClient implements ListenerProvider {
...
+      // FIXME use internal part of the offline engine for this
+      OfflineMutationsHandler.markAsOffline(mutationOptions);
```
StephenCoady: I think this function should instead take the operation. We then use setContext on this operation. As it is the function takes a context object but we can't mutate `context` from `mutationOptions`.

StephenCoady: To be honest we should probably just use `setContext` directly from within this function.

wtrocki: We are not in links. The operation was not created. Our helpers also work directly on the context because this is exactly before we passing that to mutate function. We can execute any operation on this object.

StephenCoady: Ah yes, my bad - Friday evening hiccup!

wtrocki:
Playing with this more. It looks that change has additional side impact on the flow. 
It works in showcase but integration tests shows the difference. Need 1h to take look

StephenCoady:
> Playing with this more. It looks that change has additional side impact on the flow.
> It works in showcase but integration tests shows the difference. Need 1h to take look

Maybe I am doing something wrong but I can't get this to work. I found one problem - I'll highlight the line now.

packages/offix-client/src/OfflineClient.ts
```diff
@@ -100,21 +101,28 @@ export class OfflineClient implements ListenerProvider {
...
     options: MutationHelperOptions<T, TVariables>): Promise<FetchResult<T>> {
     if (!this.apolloClient) {
       throw new Error("Apollo offline client not initialised before mutation called.");
+    }
+
+    const mutationOptions = createMutationOptions<T, TVariables>(options);
+    const isOffline = await this.config.networkStatus.isOffline();
+    if (isOffline) {
+      // FIXME use internal part of the offline engine for this
+      OfflineMutationsHandler.markAsOffline(mutationOptions);
```
StephenCoady: In this line you pass the whole operation vars but only operate on the context in OfflineMutationsHandler, I think this line should be mutationOptions.context

wtrocki:
Additional check - we need 4to stop storing operations and move to store additional elements:

- Context
- Variables
- Document

wtrocki:
Outdated

PR #130, opened 2019-07-22
fix: CacheHelpers: make update query optional

wtrocki:


wtrocki:
@darahayes Actually integration tests are checking this case many times. 
This is just because we do not have typescript in integration tests so this error did not appear.

PR #137, opened 2019-07-29
Do not require cache updates without app restart

wtrocki:
## Motivation

Global cache updates are required now to see optimistic response on every offline mutation. This change will allow us to use context attached cache update. 

Documentation is still valid, this will be just a major usability feature that will let us to experiment with sample apps without unexpected behavior.

### DIFF
--- a/packages/offix-client/src/OfflineClient.ts
+++ b/packages/offix-client/src/OfflineClient.ts
@@ -105,10 +105,10 @@ export class OfflineClient implements ListenerProvider {
     if (!this.apolloClient) {
       throw new Error("Apollo offline client not initialised before mutation called.");
     } else {
-      const helperOptions = createMutationOptions<T, TVariables>(options);
-      helperOptions.context.updateFunction = helperOptions.update;
+      const mutationOptions = createMutationOptions<T, TVariables>(options);
+      mutationOptions.context.updateFunction = mutationOptions.update;
       return this.apolloClient.mutate<T, TVariables>(
-        helperOptions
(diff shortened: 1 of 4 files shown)


packages/offix-client/src/OfflineClient.ts
```diff
@@ -105,8 +105,10 @@ export class OfflineClient implements ListenerProvider {
     if (!this.apolloClient) {
       throw new Error("Apollo offline client not initialised before mutation called.");
     } else {
+      const helperOptions = createMutationOptions<T, TVariables>(options);
+      helperOptions.context.updateFunction = helperOptions.update;
```
darahayes: Questions:

1. Naming??? `helperOptions` seems like it's come out of nowhere. Perhaps `mutationOptions`?
2. Is there a reason why we set `helperOptions.context.updateFunction` here and not internally within `createMutationOptions`?

StephenCoady: I think its better to leave it here, and then its up to the developer how they want to use it from the cache package.

If we do it in `createMutationOptions` we would be doing this:

```
options.update = update;
options.context = { ...context, returnType, idField, update };
return options;
```

wtrocki: This change is just a quick way to fix current inconvenience and enable quick getting started on sample apps. Once this is merged I will instantly focus on #125 which will require much more testing. 

This is very quick usability change for offline package. Adding it to cache package will introduce breaking change that is not desired.

PR #142, opened 2019-08-06
fix: catch errors when adding to offline queue

StephenCoady:
### Description

When adding items to the offline queue we currently do nothing if an error occurs. This can happen for a number of reasons. One being if the storage is not ready or if the cache quota is exceeded.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


StephenCoady:
Going to hold off and catch some more promises.

PR #143, opened 2019-08-06
feat: add httpOptions to init()

MStokluska:
We have implemented Offix to our chat-app and problem we have is that Offix does not support credentials in http links. Below are our initial proposed changes. Let us know what do you think!

### DIFF
--- a/packages/offix-client/src/OfflineClient.ts
+++ b/packages/offix-client/src/OfflineClient.ts
@@ -12,8 +12,9 @@ import {
 } from "offix-offline";
 import { ApolloOfflineClient } from "./ApolloOfflineClient";
 import { MutationHelperOptions, createMutationOptions } from "offix-cache";
-import { FetchResult } from "apollo-link";
+import { ApolloLink,FetchResult } from "apollo-link";
 import { buildCachePersistence } from "./cache";
+import { HttpLink } from "apollo-link-http";
 
 /**
 * Factory for creating Apollo Offline Client
@@ -60,13 +61,12 @@ export class OfflineClient implements ListenerProvider {
   /**
   * Initialize client
   */
-  public async init(): Promise<ApolloOfflineClient> {
+  public async init(httpOptions?: HttpLink.Options): Promise<ApolloOfflineClient> {
     await this.store.init();
     const cache = await buildCachePersistence(this.config.cacheStorage);
     const offlineLink = await createOfflineLink(this.config, this.store);
     const conflictLink = await createConflictLink(this.config);
-    const link = await createDefaultLink(this.config, offlineLink, conflictLink, cache);
-
+    const link = await createDefaultLink(this.config, offlineLink, conflictLink, cache, httpOptions);
     const client = new ApolloClient({
       link,
       cache
--- a/packages/offix-client/src/OfflineClient.ts
+++ b/packages/offix-client/src/OfflineClient.ts
@@ -1,4 +1,4 @@
-import { ApolloClient, MutationOptions, OperationVariables } from "apollo-client";
+import { ApolloClient, OperationVariables } from "apollo-client";
 import { OffixClientConfig } from "./config";
 import { OffixDefaultConfig } from "./config/OffixDefaultConfig";
 import { createDefaultLink, createOfflineLink, createConflictLink } from "./links/LinksBuilder";
@@ -12,9 +12,8 @@ import {
 } from "offix-offline";
 import { ApolloOfflineClient } from "./ApolloOfflineClient";
 import { MutationHelperOptions, createMutationOptions } from "offix-cache";
-import { ApolloLink,FetchResult } from "apollo-link";
+import { FetchResult } from "apollo-link";
 import { buildCachePersistence } from "./cache";
-import { HttpLink } from "apollo-link-http";
 
 /**
 * Factory for creating Apollo Offline Client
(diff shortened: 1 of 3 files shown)


packages/offix-client/src/OfflineClient.ts
```diff
@@ -60,13 +61,12 @@ export class OfflineClient implements ListenerProvider {
   /**
   * Initialize client
   */
-  public async init(): Promise<ApolloOfflineClient> {
+  public async init(httpOptions?: HttpLink.Options): Promise<ApolloOfflineClient> {
```
wtrocki: No strong opinion but:
I would prefer to go with terminating link for couple reasons:

- It is just easier to migrate to offix by passing your already existing link to init rather than figuring out how to replicate those changes to offix.
- Using terminating link allows people to configure the flow with not only Http
- Using HttpLink.Options make a strong connection on Apollo lib version

The drawback of the terminating link is that some of the config parameters will be ignored.

packages/offix-client/src/OfflineClient.ts
```diff
@@ -12,8 +12,9 @@ import {
 } from "offix-offline";
 import { ApolloOfflineClient } from "./ApolloOfflineClient";
 import { MutationHelperOptions, createMutationOptions } from "offix-cache";
-import { FetchResult } from "apollo-link";
+import { ApolloLink,FetchResult } from "apollo-link";
```
StephenCoady: no need to import `ApolloLink`

packages/offix-client/src/OfflineClient.ts
```diff
@@ -60,13 +61,12 @@ export class OfflineClient implements ListenerProvider {
...
   */
-  public async init(): Promise<ApolloOfflineClient> {
+  public async init(httpOptions?: HttpLink.Options): Promise<ApolloOfflineClient> {
     await this.store.init();
     const cache = await buildCachePersistence(this.config.cacheStorage);
     const offlineLink = await createOfflineLink(this.config, this.store);
     const conflictLink = await createConflictLink(this.config);
-    const link = await createDefaultLink(this.config, offlineLink, conflictLink, cache);
-
+    const link = await createDefaultLink(this.config, offlineLink, conflictLink, cache, httpOptions);
```
StephenCoady: This is optional in your `init` method so you will need to check it exists before passing it here or make it optional on the methods that use it.

packages/offix-client/src/OfflineClient.ts
StephenCoady: @darahayes @wtrocki If we are saying we are going to move to passing terminating links then I don't think we should break the api here. It would make more sense to add a `httpOptions` key to the config and use it from there. Otherwise, we will be making two destructive changes one after the other.

packages/offix-client/src/links/LinksBuilder.ts
```diff
@@ -81,21 +81,22 @@ export const defaultHttpLinks = async (config: OffixClientConfig, offlineLink: A
...
+  
   if (config.authContextProvider) {
     links.push(createAuthLink(config));
   }
   const localFilterLink = new LocalDirectiveFilterLink();
   links.push(localFilterLink);
 
+  const httpLinkOptions = {
+    uri: config.httpUrl,
+    ...httpOptions
```
StephenCoady: bug here. httpOptions also has a `uri` key.

wtrocki:
Summary of our talk this morning:

There is bigger picture here - we have HTTP LINK with WEBSOCKET LINK and FILE UPLOAD link. Overal complexity of this and bundles side causing real chalenge to maintain config. Matrix of options are hard to predict. For example 
if fileUpload===true Upload link will be used that might have different options that Apollo client etc. So having terminating link will be the most desirable option. For the time being we can go with the config havbing http options but I see this as only temporary solution before we introduce terminating link and fix problem in the source.

packages/offix-client/src/links/LinksBuilder.ts
wtrocki: Yeah.. Moving this to config will mean that we will have 4 places that will modify behaviour of the entire client significantly. Documenting order and priority here may be challenging. 

- config.httpOptions - contains full scope of options
- config.url = specifies url to be used.
- config.fileUpload = true (use upload link instead of HttpLink - compatibility)
- config.wsURL (use websockets on top of http)

packages/offix-client/src/config/OffixClientConfig.ts
```diff
@@ -7,6 +7,11 @@ import { ObjectState } from "offix-offline";
 import { ConflictListener } from "offix-offline";
 import { CacheUpdates } from "offix-cache";
 import { RetryLink } from "apollo-link-retry";
+import { HttpLink } from "apollo-link-http";
+
+export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
+export type HttpLinkOptions = Omit<HttpLink.Options, "uri">;
```
wtrocki: This is just on typescript level. JS side will still accept uri.
I think we just need to pass it but simply force assign url from config.uri after.

packages/offix-client/src/links/LinksBuilder.ts
```diff
@@ -88,14 +88,15 @@ export const defaultHttpLinks = async (config: OffixClientConfig, offlineLink: A
...
+  const httpLinkOptions = {
+    uri: config.httpUrl,
+    ...config.httpLinkOptions
+  };
+
   if (config.fileUpload) {
-    links.push(createUploadLink({
-      uri: config.httpUrl
-    }));
+    links.push(createUploadLink(httpLinkOptions));
```
wtrocki: At some point, options might diverge: https://github.com/jaydenseric/apollo-upload-client#function-createuploadlink

It is ok for now.

wtrocki:
@MStokluska @milenazuccarelli Do you mind creating follow up issue to use terminating link. Like this one but there is still some technical debt that we can fix. I'm happy to do follow up issue or we can collaborate to do it together.

packages/offix-client/src/config/OffixClientConfig.ts
StephenCoady: if you just swap the order of the spread operator you will get this result.

packages/offix-client/src/OfflineClient.ts
MStokluska: true, that was part of the initial plan to resolve the issue we had, thanks!

PR #146, opened 2019-08-07
feat: initial react-offix-hooks prototype

darahayes:
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/react-offix-hooks/README.md
+++ b/packages/react-offix-hooks/README.md
@@ -45,10 +45,7 @@ const App = () => {
 
   // initialize the offix client and set the apollo client
   useEffect(() => {
-    offixClient.init().then((client) => {
-      console.log('offline client initialized')
-      setApolloClient(client)
-    })
+    offixClient.init().then(setApolloClient)
   }, [])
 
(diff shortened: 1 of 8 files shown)


packages/react-offix-hooks/README.md
```diff
@@ -0,0 +1,7 @@
+## react-offix-hooks
+
+Use `offix-client` in React hooks.
```
wtrocki: Simple usage will be cool.
We can add a note that this is still experimental

packages/react-offix-hooks/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-offix-hooks",
```
wtrocki: what will be your approach for react.. helpers or entire offix-react wrapper? name can change?

darahayes: I see this module as being really small and only covering the hooks. I don't think the name will need to change. For now react integrations should/will be delivered by existing libraries out there.

PR #148, opened 2019-08-07
fix: use optimistic data if available. add client to resolve.

StephenCoady:
we need to use the parent optimistic data when available as this is most
likely what the user will be looking at when interacting with an object.
also, the client should be added to the resolve function for the case
where client diff is empty because of an optimistic response.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-offline/src/utils/cacheHelper.ts
+++ b/packages/offix-offline/src/utils/cacheHelper.ts
@@ -12,9 +12,11 @@ export const getObjectFromCache = (operation: Operation, id: string) => {
 
     if (context.cache && context.cache.data) {
       const idKey = context.getCacheKey({ __typename: context.returnType, id });
-      const optimisticData = context.cache.optimisticData.parent.data;
-      if (idKey && optimisticData[idKey]) {
-        return Object.assign({}, optimisticData[idKey]);
+      if (context.cache.optimisticData && context.cache.optimisticData.parent) {
+        const optimisticData = context.cache.optimisticData.parent.data;
+        if (idKey && optimisticData[idKey]) {
+          return Object.assign({}, optimisticData[idKey]);
+        }
       }
       const cacheData = context.cache.data.data;
       if (idKey && cacheData[idKey]) {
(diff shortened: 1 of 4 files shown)


StephenCoady:
ignore this. I don't have it solved yet.

packages/offix-offline/src/utils/cacheHelper.ts
```diff
@@ -11,8 +11,14 @@ export const getObjectFromCache = (operation: Operation, id: string) => {
     const context = operation.getContext();
 
     if (context.cache && context.cache.data) {
-      const cacheData = context.cache.data.data;
       const idKey = context.getCacheKey({ __typename: context.returnType, id });
+      if (context.cache.optimisticData && context.cache.optimisticData.parent) {
+        const optimisticData = context.cache.optimisticData.parent.data;
+        if (idKey && optimisticData[idKey]) {
+          return Object.assign({}, optimisticData[idKey]);
```
wtrocki: any reason why we create new object? This is going to ignore nested elements anyway.

wtrocki:
> fix: stop using grandparent

wtrocki:
Build needs work

packages/offix-offline/src/utils/cacheHelper.ts
```diff
@@ -10,9 +10,18 @@ const logger = debug.default("Offix:Helpers");
 export const getObjectFromCache = (operation: Operation, id: string) => {
     const context = operation.getContext();
 
-    if (context.cache && context.cache.data) {
-      const cacheData = context.cache.data.data;
+    if (context.cache && context.cache.data && !context.isOffline) {
       const idKey = context.getCacheKey({ __typename: context.returnType, id });
+      if (context.cache.optimisticData && context.cache.optimisticData.parent) {
+        let optimisticData;
+        if (context.cache.optimisticData.parent) {
```
wtrocki: I did see this before... (Line 15)

wtrocki:
@StephenCoady Linked this to showcase application. It is really fixing all the issues we have seen! Really good work!

Issue #145, opened 2019-08-07
Page fails to load online after offline refresh: item.operation.getContext is not a function


milenazuccarelli:
## Bug Report

Hey guys, I am experiencing the following error: 
`Uncaught (in promise) TypeError: item.operation.getContext is not a function`

After doing the following:
1. Load my app
2. Go offline
3. Make some changes offline
4. Refresh
5. Go back online
6. Page fails to load

<img width="735" alt="Screen Shot 2019-08-07 at 12 28 31" src="https://user-images.githubusercontent.com/24867345/62620191-8a27f300-b910-11e9-83a0-0ca76e7c42eb.png">

* **Affected Package(s)**:
    offix-client
* **Package version(s)**:
    offix-client: 0.7.1

MStokluska:
Hi @milenazuccarelli , which changes are you causing this error? Is this addMessage mutations from the app?
Also, is refreshing the page trying to refresh the link with server, which in this case would not be accessible as you are offline? I might be getting this wrong but isn't offix supporting offline as it stores mutations for later execution ( once u go back online ) while refreshing website requests response from server?

wtrocki:
This is a critical bug introduced recently. 
Going to take look

PR #151, opened 2019-08-07
Check if dealing with JSON or Operation object in persistent queue

wtrocki:
OfflineMutationsHandler will deal with two types of elements:

- Elements enqueued instantly after scheduling
- Element enqueued after restart

Second option is missing function calls and we should always check if they exist.

## Verification

1. Go offline
2. Add some tasks
3. Restart website when online 
4. See no error from #145


wtrocki:
Integration tests already covering this case.

PR #152, opened 2019-08-07
BREAKING: Enable terminating link support

wtrocki:
## Motivation

Move concept form #143 to new level by using the actual link.
Remove very heavy dependencies that reduced bundle size by 31%
Give users flexibility to pass whatever they already have in their apps.
Support plain and simple http link for backward compatibility.
Add theoretical ability to use offix-client in js-sdk (now we can pass extra stuff as links)


docs/ref-release-notes.md
```diff
@@ -12,11 +12,15 @@
 
 # Changed
```
wtrocki: I really really dislike this format @darahayes 
This is very vague and unredable. When I created it initially I was hoping to get something like: https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript

In most of the cases people will be looking into introduction for new features. This format do not support long descriptions and large migration text etc.

darahayes: If it's not working, let's change it.

wtrocki: I want to discuss about it or see examples because I'm probably missing the concept. Do you still have website that explains it ?

darahayes:
@wtrocki is there any strong reason why we pass the terminating link into `init()` As opposed to the config going into `OfflineClient`?

I kinda see this as two items of configuration that could easily be one.

I also see it reducing flexibility in how the client can be initialised and possibly leading to unnecessary boilerplate for setting up the client in user applications. 

I'm thinking a little further down the line where there could be some integration modules that might accept a fully configured `OfflineClient` and they would call `init()` under the hood for you and provide it within your application context. Passing the terminating link separately would make this a little more difficult IMO.

PR #157, opened 2019-08-09
chore: release 0.8.2

wtrocki:


StephenCoady:
@wtrocki because of the previous PR the offix-client will be old in the hooks package now. I know its the same package but no harm in changing it now so it doesn't get missed next time.

PR #165, opened 2019-09-03
Initial version of the scheduler refactor

wtrocki:
The target of this PR is to decouple Scheduling logic from Apollo link 

For testing use: https://github.com/aerogear/ionic-showcase/pull/253
execute npm link on this repository and then link it to showcase.


wtrocki:
👁

packages/offix-client/src/OfflineClient.ts
```diff
@@ -135,14 +175,22 @@ export class OfflineClient implements ListenerProvider {
...
       this.apolloClient as ApolloOfflineClient,
       this.config);
     offlineLink.setup(offlineMutationHandler);
     // Reschedule offline mutations for new client instance
     await offlineMutationHandler.replayOfflineMutations();
+
     // After pushing all online changes check and set network status
-    await offlineLink.initOnlineState();
+
+    // ^^ But why do we wait until now to check and set network status?
```
wtrocki: Simply because if we let offline changes go without enqueueing then they will be:

- Non ordered (edit can be send faster than creation)
- Non retried (there is no protection for online changes now)
- Not going thru offline processor so no notifications etc.

We just wait to enqueue all offline changes and then initialize state. 
This means if we are offline - nothing happens if we are online offline scheduler will kick in and start replicating those. This is not perfect but it is the only way to avoid some more messy issues.

packages/offix-client/src/OfflineClient.ts
```diff
@@ -135,14 +175,22 @@ export class OfflineClient implements ListenerProvider {
...
+    await offlineLink.initOnlineState(); // TODO this needs to go away
```
wtrocki: It will be hard for this to go away because the changes will be schedulerd. 
Offline `this.offlineProcessor.initOnlineState();` can go away as it is being used only as inidicator and do not have any mutations waiting. Hopefuly we can bring all of those outside link at some point and.. then both will go away :D

darahayes:
Closing this PR to open a new one

darahayes:
Closed in favour of https://github.com/aerogear/offix/pull/170

PR #166, opened 2019-09-06
removed operation from persistence

StephenCoady:
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-client/src/OfflineClient.ts
+++ b/packages/offix-client/src/OfflineClient.ts
@@ -122,7 +122,6 @@ export class OfflineClient implements ListenerProvider {
     if (!this.apolloClient) {
       throw new Error("Apollo offline client not initialised before mutation called.");
     } else {
-      debugger
       const mutationOptions = createMutationOptions<T, TVariables>(options);
       mutationOptions.context.cache = this.apolloClient.cache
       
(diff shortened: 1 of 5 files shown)


packages/offix-client/src/OfflineClient.ts
```diff
@@ -135,13 +134,13 @@ export class OfflineClient implements ListenerProvider {
         variables: mutationOptions.variables
       })
 
+      // TODO we could probably move the conflictBase directly on to the mutationOptions.
+      // does createMutationOptions living in offix-cache make sense then though?
```
wtrocki: Generally, the side effects of the refactor to the scheduler are that people will not need to bother with Apollo cache. Apollo proposes to use graphql (and InMemoryCache) for a persisting state. I think people want to simply have state available (and persisted) - what I mean for us is that optimistic responses can become just optional feature that people can add (along with update methods) not the element on the critical path. 
Other offline providers use the same approach

wtrocki: TL;DR conflict base in mutationOptions will be against responsibility. Our base will be taken once someone will schedule offline operation

packages/offix-offline/src/offline/OperationQueueEntry.ts
```diff
@@ -58,15 +64,17 @@ export class OperationQueueEntry implements OfflineItem {
...
 
   /**
    * Adapt object in order to persist required information
    */
   public toOfflineItem(): OfflineItem {
     return {
-      operation: this.operation,
+      query: this.query,
+      variables: this.variables,
+      operationName: this.operationName,
```
darahayes: Question: do we/will we need the `context` in here?

StephenCoady:
merged

PR #183, opened 2019-10-03
feat: apollo specific types for generics like OfflineQueue & OfflineStore

darahayes:
I have a dev release with these changes published. You can get it at with `npm install offix-client@next`

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/offix-client/src/OfflineClient.ts
+++ b/packages/offix-client/src/OfflineClient.ts
@@ -1,14 +1,20 @@
 import { ApolloClient, OperationVariables, MutationOptions } from "apollo-client";
 import { OffixClientConfig } from "./config/OffixClientConfig";
 import { OffixDefaultConfig } from "./config/OffixDefaultConfig";
-import { ApolloOperationSerializer } from "./apollo/ApolloOperationSerializer";
-import { createCompositeLink } from "./apollo/LinksBuilder";
-import { ApolloOfflineClient } from "./apollo/ApolloOfflineClient";
 import {
+  ApolloOfflineQueue,
+  ApolloOfflineStore,
+  ApolloOfflineQueueListener,
+  ApolloQueueEntryOperation,
+  ApolloIResultProcessor,
+  ApolloOperationSerializer,
+  ApolloOfflineClient,
+  createCompositeLink,
   addOptimisticResponse,
   removeOptimisticResponse,
   restoreOptimisticResponse
-} from "./apollo/optimisticResponseHelpers";
+} from "./apollo/";
+
 import {
   ApolloCacheWithData,
   OfflineStore,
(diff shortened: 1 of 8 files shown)


wtrocki:
like next label! good idea

packages/offix-client/src/OfflineClient.ts
```diff
@@ -141,22 +147,22 @@ export class OfflineClient {
     // Optimistic Responses
     // TODO move this somewhere that makes sense, also handle error cases
     this.queue.registerOfflineQueueListener({
-      onOperationEnqueued: (operation: QueueEntryOperation<MutationOptions>) => {
+      onOperationEnqueued: (operation: ApolloQueueEntryOperation) => {
```
wtrocki: ok ok.. but we wanted to get generic later on. 
This will straightaway mean that we going to break stuff. 
Any reason why this needs to be apollo?

wtrocki:
This is my fault I kinda misinterpreted this as we will declare the types for the end users but do not hange our internals so it will be still generic enough. Let's chat about it as there are couple ways here.

packages/offix-client/src/OfflineClient.ts
darahayes: I don't really see what's the problem here. OfflineClient is a specific implementation of the generic Queue + Store mechanisms.

darahayes: Later down the line we will have to untangle some small amount of logic from the `OfflineClient` and that logic + generic queue + generic store will become a scheduler package.

At that point OfflineClient will simply become a concrete implementation of the scheduler using all ApolloSpecific types. Does that make sense?

darahayes: The Queue and Store itself are still completely generic though. IMO there is no need for a user of `offix-client` to care about those being generic because they are using the very Apollo specific bundle.

The main benefit of the generics is for maintenance and for advanced cases like implementation of new schedulers (e.g. rest/custom)

wtrocki: Ok. I missed that we want OfflineClient to be Apollo specific. It makes a lots of sense from many aspects including compatibility and support.

packages/offix-client/src/apollo/ApolloOfflineClient.ts
```diff
@@ -1,8 +1,22 @@
...
+  IResultProcessor,
+  QueueEntryOperation,
+  QueueEntry
+} from "offix-offline";
+import { MutationHelperOptions } from "offix-cache";
+
+export type ApolloOfflineQueue = OfflineQueue<MutationOptions>;
+export type ApolloOfflineStore = OfflineStore<MutationOptions>;
+export type ApolloOfflineQueueListener = OfflineQueueListener<MutationOptions>;
+export type ApolloIResultProcessor = IResultProcessor<MutationOptions>;
```
wtrocki: ApolloOfflineResultProcessor?

packages/offix-client/src/apollo/ApolloOfflineClient.ts
```diff
@@ -1,8 +1,22 @@
...
+  OfflineStore,
+  OfflineQueueListener,
+  OfflineQueue,
+  IResultProcessor,
+  QueueEntryOperation,
+  QueueEntry
+} from "offix-offline";
+import { MutationHelperOptions } from "offix-cache";
+
+export type ApolloOfflineQueue = OfflineQueue<MutationOptions>;
```
wtrocki: It will be cool to get some minimum docs for those.

packages/offix-client/src/apollo/ApolloOfflineClient.ts
wtrocki: Ignore me here.

packages/offix-client/src/apollo/ApolloOfflineClient.ts
darahayes: Okay going to do that before merging

PR #226, opened 2019-11-01
doc: updates for 0.9.2 release

darahayes:
##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


docs/getting-started.md
```diff
@@ -22,26 +22,64 @@ yarn add offix-client
...
 import { OfflineClient } from 'offix-client';
-```
 
-## Configuration
+const config = {
+  httpUrl: 'http://localhost:4000/graphql'
+};
 
-To provide custom configuration to the client, the following options are available. If you wish, these are also available by using the `OffixClientConfig` interface from the SDK.
+// offlineClient is a parent around the apollo client
```
wtrocki: What this means?

darahayes: Sorry maybe bad language. `OfflineClient` class really is a container/wrapper that has the extended apolloClient as a property

darahayes: Not really sure how to get that across

wtrocki: Ah.. right. Sorry it is hard to actually understand this sentence. Maybe 2 sentences?

docs/getting-started.md
```diff
@@ -22,26 +22,64 @@ yarn add offix-client
...
+  variables: {
+    name: 'hello world!'
+  }
 };
+
+client.offlineMutate(options).catch((error) => {
+  // we are offline - lets wait for changes
+  if(error.offline) {
+    error.watchOfflineChange().then((result) => {
+      console.log('mutation was completed after we came back online!', result)
```
wtrocki: I really like how simple it is now!

docs/getting-started.md
```diff
@@ -22,26 +22,64 @@ yarn add offix-client
...
-To provide custom configuration to the client, the following options are available. If you wish, these are also available by using the `OffixClientConfig` interface from the SDK.
+// offlineClient is a parent around the apollo client
+const offlineClient = new OfflineClient(config);
 
-```javascript
-let config = {
-  httpUrl: 'http://localhost:4000/graphql',
-  wsUrl: 'ws://localhost:4000/graphql'
+// client is the initialized apollo client
+const client = await client.init();
```
wtrocki: In future we will need to probably give more react/angular examples with progressbar

PR #241, opened 2019-11-11
fix: OfflineClient uses offix-scheduler under the hood

darahayes:
##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


packages/offix-client/src/config/OffixClientConfig.ts
```diff
@@ -1,99 +1,55 @@
...
-
-  /**
-   * [Modifier]
-   *
-   * The options to configure how failed offline mutations are retried.
-   *
-   */
-  retryOptions?: RetryLink.Options;
+export class OffixClientConfig implements OffixClientOptions {
+  public httpUrl?: string;
```
wtrocki: That is breaking change. Any reason why we have done that :) ?

packages/offix-client/src/OfflineClient.ts
```diff
@@ -121,15 +103,24 @@ export class OfflineClient {
         debug: false
       });
     }
+
+    this.offix = new Offix({
```
wtrocki: I do not like name of this. Mentioned it in different PR. Offix is vague. OffixScheduler works much better.

packages/offix-client/src/config/OffixClientConfig.ts
```diff
@@ -1,99 +1,55 @@
...
+  public conflictProvider = new VersionedState();
+  public networkStatus?: NetworkStatus;
+  public terminatingLink: ApolloLink | undefined;
+  public cacheStorage?: PersistentStore<PersistedData>;
+  public offlineStorage?: PersistentStore<PersistedData>;
+  public conflictListener?: ConflictListener;
+  public mutationCacheUpdates?: CacheUpdates;
+  public cache?: InMemoryCache;
+
+  public retryOptions = {
```
wtrocki: [Question] Do we actually still support retries (retry link) after moving to scheduler?

packages/offix-client/src/config/OffixClientConfig.ts
darahayes: This looks like a breaking change but it's actually not. We used to have two files. 

* OffixClientConfig - the interface that defined what options can be passed into the OfflineClient constructor by a user.
* OffixDefaultConfig - actual config class that we instantiate passing in user provided options and merging some defaults

I felt that the names were always really confusing and made little sense. So now we have the following:

* OffixClientOptions - Options represents the things a user can pass in
* OffixClientConfig - Config represents an actual instance of the config.

Because I renamed the files, it looks like I changed the options users can pass in but you'll actually find them all in `OffixClientOptions.ts` and there have been no changes at all.

packages/offix-client/src/config/OffixClientConfig.ts
darahayes: We do in OfflineClient using the Apollo RetryLink. The scheduler doesn't have it though. It's something we should definitely be looking at.

packages/offix-client/src/config/OffixClientConfig.ts
wtrocki: Perfect!

packages/offix-client/package.json
```diff
@@ -43,12 +43,15 @@
...
+    "apollo-link-error": "^1.1.12",
     "apollo-link-http": "1.5.16",
     "apollo-link-retry": "2.2.15",
     "debug": "4.1.1",
     "idb-localstorage": "0.2.0",
     "offix-cache": "0.9.0",
-    "offix-offline": "0.9.0"
+    "offix-conflicts-client": "0.9.0",
+    "offix-offline": "0.9.0",
+    "offix-scheduler": "0.9.0"
```
wtrocki: Note - let's  publish new packages manually using different version and then configure proper permissions.

packages/offix-client/src/apollo/optimisticResponseHelpers.ts
```diff
@@ -30,3 +31,39 @@ export function restoreOptimisticResponse(
     addOptimisticResponse(apolloClient, { op, qid });
   }
 }
+
+/**
+ * Imagine we do a mutation that creates a new object while offline.
```
wtrocki: Strange form :)

packages/offix-scheduler/src/OffixScheduler.ts
```diff
@@ -2,19 +2,20 @@ import { OffixOptions } from "./config/OffixOptions";
...
+
+import {
+  OfflineQueue,
+  OfflineQueueListener
+} from "./queue";
+
+import { OfflineStore } from "./store";
 
 /**
  * Offix
```
wtrocki: Comment is outdated

wtrocki:
Good work! I would wait till we publish initial versions of packages to get another shot on how this will affect js-sdk and showcase

wtrocki:
Got another shot! I do think that if we merge this we will need to change our documentation a little bit. It will be cool to get some of the changes into this PR.

darahayes:
@wtrocki I've added a new section to the website so it now looks something like this. It's still really early phase and will probably evolve as I work on a more featured example (including HTTP cache etc.)

Screenshot 2019-11-18 at 12 16 12

darahayes:
@wtrocki I think this is ready for review now. I have these changes tested in the showcase app. There is a dev version of the offix packages published, which you can install with `npm install offix-client@next`

docs/ref-offix-scheduler.md
```diff
@@ -0,0 +1,111 @@
...
+id: offix-scheduler-introduction
+title: Offix Scheduler Introduction
+sidebar_label: Introduction
+---
+
+Offix Scheduler is the core component that delivers the offline queueing and storage in packages such as `offix-client`. The scheduler can be used in an application directly to build offline workflows, or it can be as the foundation to build new packages and clients with offline functionality.
+
+Unlike `offix-client` which provides many features out of the box, `offix-scheduler` provides a lower level set of core functionalities which can be used to build offline experiences.
+
+* Flexible `NetworkStatus` interface for defining when an app is and is not offline.
```
wtrocki: Flexible sounds strange

wtrocki:
I have mixed feelings about scheduler as separate docs chapter. Fine for now but I do see us moving towards platforms and supported clients and actual scheduler being an implementation detail or contributing guide

PR #256, opened 2019-11-18
refactor: use optional chaining

b1zzu:
### Description

Migrate from chain of conditions to typescript optional chaining in order to make the code more readable.

The only side effect is that VSCode seems to recognize it as an error.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


b1zzu:
@darahayes I would like to add some unit tests for the `getConflictData` method, where could I add them?

darahayes:
@b1zzu Thanks for the PR! Just a heads up, a pretty big PR was merged recently (#241) that moved this file to the `offline-client` package. There is a `test` folder in there where you can create a new file, and you can pretty much copy/paste the structure from one of the other files.

packages/offix-client/test/apollo/conflicts/ConflictLink.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      undefined,
+      undefined,
+      {
+        exception: {
+          conflictInfo: {}
+        }
+      }
+    )
+  ]);
+  expect(result).toEqual({});
```
darahayes: Nice test cases. Could this maybe be some other value that makes it more obvious that it's meant to be the `exception.conflictInfo` property. It is not immediately obvious that's what we're matching. Even just some silly string value or maybe add some key/values so it's not an empty object.

b1zzu:
@darahayes the problem with the current errors in the ci is this typescript option: `"noImplicitAny": true` I think there should be a way to disable it for test files

packages/offix-client/test/apollo/conflicts/ConflictLink.test.ts
b1zzu: I've added the full object

wtrocki:
+1 for disabling strict mode for tests

wtrocki:
Actually I have linked that and we cannot use it for the cases when someone would want to use our project from typescript - we do that and chaining requires latest typescvript that suddenly clashes with angular tooling

b1zzu:
@wtrocki then I can revert the optional chaining but we can keep the tests

wtrocki:
That will be perfect. I like optional chaining but as new things come up there is need for polyfil for old typescript and not sure if this is worth it.

packages/offix-client/src/apollo/conflicts/ConflictLink.ts
```diff
@@ -108,7 +108,7 @@ export class ConflictLink extends ApolloLink {
   private getConflictData(graphQLErrors?: ReadonlyArray<GraphQLError>): ConflictInfo | undefined {
     if (graphQLErrors) {
       for (const err of graphQLErrors) {
-        if (err.extensions && err.extensions.exception && err.extensions.exception.conflictInfo) {
+        if (err.extensions?.exception?.conflictInfo) {
```
wtrocki:
```suggestion
        if (err.extensions && err.extensions.exception && err.extensions.exception.conflictInfo) {
```

PR #263, opened 2019-11-20
fix: Extract apollo client creation into protected function

wtrocki:
## Usage
```ts
class MyClient extends OffixClient {
  protected createClient(){
   ...
  }
} 
```

### DIFF
--- a/packages/offix-client/src/OfflineClient.ts
+++ b/packages/offix-client/src/OfflineClient.ts
@@ -136,11 +136,8 @@ export class OfflineClient {
 
     const link = await createCompositeLink(this.config, conflictLink);
 
-    const client = new ApolloClient({
-      link,
-      cache: this.cache
-    });
-
+    this.createApolloClient(link, cache);
+  
     this.apolloClient = this.decorateApolloClient(client);
 
     // Optimistic Responses
@@ -169,6 +166,17 @@ export class OfflineClient {
     });
     await this.scheduler.init();
     return this.apolloClient;
+  }
+  
+  /**
+  * Developers can extend and replace this function to inject extra parameters
+  * into the ApolloClient instance
+  */
+  protected createApolloClient(link: ApolloLink, cache: InMemoryCache){
+    const client = new ApolloClient({
+      link,
+      cache: this.cache
+    });
   }
 
   public async execute(options: MutationOptions) {


darahayes:
@wtrocki so I'm just looking at this closer and I'm thinking about how to keep the configuration API consistent and simple. Currently, the user can pass an `InMemoryCache` instance and also a `terminatingLink` into the `OfflineClient` constructor.

If we allow the user to override the `createApolloClient` function, then do we need these config options anymore? Obviously it is nice to maintain backwards compatibility but what happens when users try to use a mixture of both approaches?

I'm kinda thinking we have the following:

* Default Case - pass nothing but httpUrl - Client and Link are created automatically
* Advanced Case - user creates their own client, but they must include an InMemoryCache instance, they must concat with our composite link (RetryLink and ConflictLink). Because the `client.link` property is public, we could do some weirdness where we do something like `client.link = ConflictAndRetryLink.concat(client.link)` but this might lead to strange and unexpected behaviour.

* User passes `InMemoryCache` via config - deprecate this
* User passes `terminatingLink` via config - deprecate this

what do you think?

darahayes:
Additionally, instead of asking users to extend this function - we could allow a builder function to be passed as a config. This might be more desirable to vanilla JS users who do not like/want to be extending classes. Instead they simply pass in the function that will create and return the client.

Then we can keep the `InMemoryCache` and `terminatingLink` options hanging around but they will simply be ignored if the `createApolloClient` function is also passed in through the config. What do you think?

darahayes:
The more I think about it, the more I like the idea that the user would pass in the client creation function as a config option instead of extending the `OfflineClient` class.

wtrocki:
I think that is amazing idea. I would not even worry about deprecating this. Let's just simply do things we want and see if that works

wtrocki:
@darahayes Keep in mind we initially had linkBuilder option but we have removed it. 
I also think that we should consider extending Apollo client. 

Attaching some methods on top of generated objects was never something I was proud off.

darahayes:
@wtrocki cool I'm going to go with the idea of allowing a creation function to be passed instead of extending the `OfflineClient` class.

wtrocki:
Yes.. We need to address this problem once for a while.

darahayes:
@wtrocki closing this in favour of https://github.com/aerogear/offix/pull/270

PR #267, opened 2019-11-21
chore: removed bad/unused test files

darahayes:
### Description

These were unused duplicates that were simply causing issues. Not worth it to keep them around

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


b1zzu:
@darahayes also the `test/server/`, `test/utils/` and `test/fixtures` directories where used only by the integration tests, so we can remove them

PR #271, opened 2019-11-27
Reorganize Integration Tests

b1zzu:
### Description

I'm looking into a more clear way to organize and divide integration tests

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


b1zzu:
@darahayes I'm experimenting a new way to organize integration tests, because the existing tests inside the `test/integration` folders are broken, and integration tests are not meant to run in parallel, and if we put them in the same directory as the unit tests then we will have to run also all unit tests sequentially, wdyt?

darahayes:
Hey @b1zzu so I actually have an open PR to simply delete those https://github.com/aerogear/offix/pull/267 as they were never really working for me and I started to realise that the karma tests are actually quite useful.

In the real world Offix relies on a couple of browser APIs and it probably is wise for our tests to be working against those. The current karma integration tests do mock out certain things such as storage but I think we might actually be in a better place if we can continue browser testing and move towards using fewer mocked APIs. What do you think?

b1zzu:
@darahayes I think we should go and merge #267 

From the coverage point of view testing with Karma is better, knowing that the tests run on a real browser gives more certainty, but at the same time, different browsers implement the api in a different way so running them on only one browser is also an optimistic way of seeing it. On the other side having the tests in Jest will make them easier to debug and write (karma when reporting error is usually losing some critical logs) so we could expect to have more integration tests in the future.

I personally would take the risk and go with Jest, but it's a risk

darahayes:
> From the coverage point of view testing with Karma is better, knowing that the tests run on a real browser gives more certainty

Agreed. Offix does have some strong reliance on indexeddb/localstorage and networking APIs and there's some possibility for other browser API usage in the future so despite the fact that Jest might be a bit easier to debug/use, I think stayin with Karma is the best approach for now. As the core maintainer right now it suits me just fine!

I would be very open to investigating alternatives to karma that would also allow us to do browser testing. For example [Cypress](https://docs.cypress.io/) is gaining lots of popularity. They currently only support chrome but Firefox support is being worked on.

Are you happy for me to close this PR and fo with #267 ?

b1zzu:
I understand the reason and I agree, so yes, we can close this PR and preserve the integration tests in karma

PR #275, opened 2019-12-02
Integrate graphql-testx

b1zzu:
##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


packages/offix-client/integration_test/package.json
```diff
@@ -3,9 +3,7 @@
   "version": "1.0.0",
   "description": "Integration tests for datasync",
   "scripts": {
-    "pretest": "kill $(lsof -t -i:4001) & node server/index.js &",
-    "test": "karma start",
-    "posttest": "kill $(lsof -t -i:4001) || true"
+    "test": "gulp test"
```
darahayes: @b1zzu is there any particular reason for adding gulp here?

b1zzu: Using gulp it's possible to start the server and run the karma tests all in the same nodejs process, which means that once the tests are finished also the server will be closed. And I can also expose the server to a random port and pass it to the karma:  https://github.com/aerogear/offix/blob/41a8c1a849e657ba8dacca092a22449353f021a2/packages/offix-client/integration_test/gulpfile.js#L30 
Mine is more an experiment, we can also achieve the same result with `pretest` and `posttest` using a fixed port for the server.
Wdyt?

darahayes: I'd much prefer to go with pre and post test commands. The JavaScript community as a whole has mostly moved away from task runners like grunt and gulp as a whole because they add a lot of maintenance overhead and for little benefits. `npm` already is a perfectly good task runner that does not require any additional dependencies.

packages/offix-client/integration_test/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "1.0.0",
   "description": "Integration tests for datasync",
   "scripts": {
-    "pretest": "kill $(lsof -t -i:4001) & node server/index.js &",
+    "pretest": "kill $(lsof -t -i:4001); kill $(lsof -t -i:4002); node server/index.js & node server/testx.js &",
```
b1zzu: @darahayes I would like to initially keep both testx and the old server in order to smoothly migrate all tests, but this looks akwards for me, do you have any better idea?

packages/offix-client/integration_test/package.json
b1zzu: Sounds good for me, I've reverted to npm

wtrocki:
I love this!

packages/offix-client/integration_test/package.json
darahayes: Sounds like a good start to me!

packages/offix-client/integration_test/test/offline.test.js
```diff
@@ -51,13 +63,15 @@ describe('Offline mutations', function () {
...
   };
 
-  const optimisticOptions = {
-    mutation: ADD_TASK,
-    operationType: CacheOperation.ADD,
-    returnType: "Task",
-    variables: newTask,
-    updateQuery: GET_TASKS
-  };
+  const newOptimisticResponse = () => {
```
darahayes: questioning the need to wrap this stuff up in a function. I think we could keep that the way it was.

b1zzu: The problem has to do with the `CREATE_TASK` mutation, if I don't use a function the `optimisticOptions` will be created with `mutation: undefined` because `CREATE_TASK` is assigned in the line `92`, the alternative to the function are:
* forcing the `optimisticOptions` to hold the pointer to the `CREATE_TASK` variable
* assign the `optimisticOptions` after assigns the `CREATE_TASK`
In any case, I would like to refactor again the tests and remove queries and mutations form global variables, the less global variables we use the better it is, but I want to do it in a sperate PR

wtrocki:
Thank you @b1zzu. I have played with testx locally on side project and I'm very impressed. 
The only limitations are the subscriptions which we goijng to add to graphback in comming days.

Reviewed this PR as well. Good to merge! Amazing work!

PR #277, opened 2019-12-03
fix: Platform docs

wtrocki:
Adding very rough although highly needed information about our platform support.
This needs work howver it is good start to get offix docs into the right shape

### DIFF
--- a/docs/android.md
+++ b/docs/android.md
@@ -1,0 +1,13 @@
+---
+id: android
+title: Android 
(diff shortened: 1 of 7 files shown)


packages/offix-conflicts-server/package.json1
```diff
@@ -0,0 +1,41 @@
+{
```
wtrocki: My bad. This will be removed

Issue #298, opened 2019-12-12
[conflicts] base calculation doesn't work with custom id fields


darahayes:
## Bug Report

Based on how this code is called: https://github.com/aerogear/offix/blob/76a0ee76fa69962d4db571165fcb2a9cf65d9247/packages/offix-client/src/apollo/conflicts/baseHelpers.ts#L55

When custom a `idField` is used then the base will not be retrieved.

* **Affected Package(s)**:
* **Package version(s)**:
* **Platform (e.g. Android/iOS)**:
* **Platform Version**:
* **Cordova Version**:
* **Node.js / npm versions**:

PR #300, opened 2019-12-13
fix: pass entire mutation options and typename to dataIdFromObject

darahayes:
##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/offix-client/src/apollo/conflicts/baseHelpers.ts
+++ b/packages/offix-client/src/apollo/conflicts/baseHelpers.ts
@@ -34,12 +34,10 @@ export function getBaseStateFromCache(
   mutationOptions: MutationOptions
 ): ConflictResolutionData {
   const context = mutationOptions.context;
-  const idField: string = context.idField || "id";
-  const id: string = mutationOptions.variables && mutationOptions.variables[idField];
 
   if (!context.conflictBase) {
     // do nothing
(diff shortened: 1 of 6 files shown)


packages/offix-client/src/apollo/conflicts/baseHelpers.ts
```diff
@@ -34,12 +34,10 @@ export function getBaseStateFromCache(
   mutationOptions: MutationOptions
 ): ConflictResolutionData {
   const context = mutationOptions.context;
-  const idField: string = context.idField || "id";
```
wtrocki: This was the whole reason id field could be passed to the mutationOptions - we wanted to know how to read base info. If we removed that we no longer need that idFIeld parameter. I have checked that quickly so not 100% sure but it is worth to mention it.

darahayes: It is still used in the mechanism where we update operations in the queue that refer to optimistic ids to real ids. example: create an object while offline and do subsequent modifications.

It may be possible at a later time to remove it completely but not yet!

darahayes:
Fixes https://github.com/aerogear/offix/issues/298

darahayes:
Before merging I'm going to add some quick documentation on using the custom idField

Issue #281, opened 2019-12-04
Add persistor options in ApolloOfflineClient constructor


JaEdmuva:
## Feature Request

**Is your feature request related to a problem? Please describe.**
I'm using v0.11.0. It is not possible to specify persistor options in the `ApolloOfflineClient` constructor. When using with react native, it is necessary to use a 'wrapper' to force serialization of data saved in AsyncStorage since the default options passed to the persistor has the `serialize` option value as `false` which causes a problem when persisting the cache data.

**Describe the solution you'd like**
Instead of having to define a 'wrapper' which is just a workaround, not a solution. Allow to define optional persistor options through the ApolloOfflineClient constructor. This will even solve #273

**Describe alternatives you've considered**
None.

wtrocki:
Ah.. right. We kinda provided workaround for that. 
Please see official docs: https://offix.dev/docs/react-native

We detect if object is string or object, but that will be much better approach. 
As for persistor itself. This is the second issue like that. 
It is possible to provide persistor (using undocumented api).
See here: https://github.com/aerogear/offix/issues/273

So TL;DR 

1. We going to fix it! 
2. It is possible to make it work anyway by following the docs example
3. You can also provide your own persitor (documented in issue above)

JaEdmuva:
@wtrocki about item #3. Yes, I can provide my own persistor, in fact I do. But there is an issue with that, which I do not fully understand yet. The thing is that when you first create the ApolloOfflineClient, you create a persistor with `serialize: false` and It doesn't matter if I later define my own, it is already created. 
Then (and it's my guess) the time comes when the first persistor still tries to read/write the cache to AsyncStorage and the exception comes, since it's trying to write an object (array) to the store but it only supports Strings. That'll be the main issue with RN I think.

darahayes:
@JaEdmuva we'll definitely work on allowing you to pass down the persistor options  through the constructor. In the meantime did you try following the steps for passing in a custom storage mechanism in our react native document? This storage mechanism might help you get past the error you are seeing. See https://offix.dev/docs/react-native#integration

darahayes:
@JaEdmuva we recently merged and released https://github.com/aerogear/offix/pull/306 which allows you to create your own `CachePersistor` object and pass it into the `OfflineClient` constructor. You can check the PR description or [the client configuration docs](https://offix.dev/docs/client-configuration#cachepersistor) to learn how to do it. It can be used with version `0.13.1`

darahayes:
[...]

Issue #273, opened 2019-12-01
How to set apollo-cache-persist options like maxSize, debug, debounce, trigger


kunaldodiya:
How to set apollo-cache-persist options like maxSize, debug, debounce, trigger in react native app ?

wtrocki:
Sadly there is no public api for that but it is very easy to do it. 

Persistor is public so.. you can set new Persistor with custom options:
https://github.com/aerogear/offix/blob/master/packages/offix-client/src/ApolloOfflineClient.ts#L52-L58

Swapping persistor needs to happen before calling init function as it is used there:
https://github.com/aerogear/offix/blob/master/packages/offix-client/src/ApolloOfflineClient.ts#L73-L76

kunaldodiya:
Thanks, I will try and let you know

wtrocki:
@darahayes IMHO we might need to document this issue

darahayes:
@kunaldodiya we recently merged and released https://github.com/aerogear/offix/pull/306 (in version `0.13.1`) which allows you to create your own `CachePersistor` object and pass it into the `OfflineClient` constructor. This should let you set whatever options you require for cache persistence. You can check the PR description or [the client configuration docs](https://offix.dev/docs/client-configuration#cachepersistor) to learn how to do it. I'm going to close this issue now but please do reach out if you have any further questions/feedback/issues.

PR #306, opened 2019-12-16
Presistor options

darahayes:
### Description

Allows the developer to pass their own `CachePersistor` object. This would close both https://github.com/aerogear/offix/issues/281 and https://github.com/aerogear/offix/issues/273

It can be done as follows:

```js
import { ApolloOfflineClient, createDefaultCacheStorage } from "offix-client";
import { HttpLink } from "apollo-link-http";
import { InMemoryCache } from "apollo-cache-inmemory";
import { CachePersistor } from "apollo-cache-persist";

const link = new HttpLink({ uri: "http://example.com/graphql" });
const cache = new InMemoryCache()

const cachePersistor = new CachePersistor({
  cache,
  storage: createDefaultCacheStorage()
})

const client = new ApolloOfflineClient({
  cache,
  cachePersistor,
  link
});
```

One thing to note, now that users can pass their own cachePersistor object. Do we need the `cacheStorage` option anymore. If you pass your own CachePersistor object then the `cacheStorage` option is no longer needed. I didn't remove it yet though because it would be a breaking change. I am happy to keep both options there but I'd like to get feedback. Any thoughts @wtrocki?

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/docs/ref-configuration.md
+++ b/docs/ref-configuration.md
@@ -39,7 +39,7 @@ const client = new ApolloOfflineClient({
 });
 ```
 
-Note: if using TypeScript, you may need to declare the cachePersistor as follows `const cachePersistor = new CachePersistor<NormalizedCacheObject>(...options)` or you may experience compiler errors.
+Note: if using TypeScript, you may need to declare the cachePersistor as follows `const cachePersistor = new CachePersistor<object>(...options)` or you may experience compiler errors.
 
 This example uses `createDefaultCacheStorage` to create the default IndexedDB based storage driver. 
 The storage can be swapped depending on the platform. For example `window.localstorage` in older browsers or `AsyncStorage` in [React Native](./react-native.md).
--- a/docs/ref-configuration.md
+++ b/docs/ref-configuration.md
@@ -12,14 +12,67 @@ sidebar_label: Client Configuration
 
 There are some additional options specific to `ApolloOfflineClient`.
 
-* `cacheStorage` - The [PersistentStore](https://github.com/aerogear/offix/blob/master/packages/offix-scheduler/src/store/PersistentStore.ts) you want your client to use for persisting the Apollo Cache (Uses indexeddb by default).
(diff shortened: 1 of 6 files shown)


docs/ref-configuration.md
```diff
@@ -12,14 +12,69 @@ sidebar_label: Client Configuration
 
 There are some additional options specific to `ApolloOfflineClient`.
 
-* `cacheStorage` - The [PersistentStore](https://github.com/aerogear/offix/blob/master/packages/offix-scheduler/src/store/PersistentStore.ts) you want your client to use for persisting the Apollo Cache (Uses indexeddb by default).
```
wtrocki: What happened here?

docs/ref-configuration.md
```diff
@@ -12,14 +12,69 @@ sidebar_label: Client Configuration
...
-* `conflictProvider` - [ObjectState](./ref-conflict-server.md#implementing-custom-conflict-resolution) Interface that defines how object state is progressed. This interface needs to match state provider supplied on server.
-* `conflictStrategy` - [ConflictResolutionStrategy](https://github.com/aerogear/offix/blob/master/packages/offix-conflicts-client/src/strategies/ConflictResolutionStrategy.ts)interface used on the client to resolve conflicts. The [default strategy](https://github.com/aerogear/offix/blob/master/packages/offix-conflicts-client/src/strategies/strategies.ts) merges client changes onto the server changes.
-* `mutationCacheUpdates` - [CacheUpdates](./ref-offline.md#global-update-functions) Cache updates functions for your mutations. Argument allows to restore optimistic responses on application restarts.
-* `retryOptions` - The options to configure how failed offline mutations are retried. See [`apollo-link-retry`](https://www.apollographql.com/docs/link/links/retry/).
+#### `cachePersistor`
```
wtrocki: I guess we can put that into the bottom as it is not the most important flag

PR #337, opened 2020-01-22
fix: error handling for when indexedDB not available

kingsleyzissou:
### Description
- Implemented error handling for when indexedDB not available (e.g. Firefox private mode), preventing the app from crashing
- Removed idb-localstorage dependency and added the class directly to offix-scheduler as IDBLocalStore
- Added tess for offix when indexedDB not available and added tests for IDBLocalStore

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/offix-scheduler/test/IDBLocalStore.test.ts
+++ b/packages/offix-scheduler/test/IDBLocalStore.test.ts
@@ -1,0 +1,79 @@
+import { IDBLocalStore } from "../src/store";
+import "fake-indexeddb/auto";
+
+let store: IDBLocalStore;
+
+describe("Empty store tests", () => {
+
+    beforeEach(() => {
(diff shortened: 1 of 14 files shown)


packages/offix-scheduler/test/IDBLocalStore.test.ts
```diff
@@ -0,0 +1,76 @@
+import { IDBLocalStore } from "../src/store";
+import "fake-indexeddb/auto";
+
+let store: IDBLocalStore;
+
+describe("Empty store tests", () => {
+
+    beforeEach(() => {
```
darahayes: Can you please change this file to use 2 spaces instead of 4 for tab width?

PR #345, opened 2020-01-28
feat: add getting started example app for React

kingsleyzissou:
### Description
- Added initial example app for getting started in React.
- Basic instructions for getting started in the README

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [ ] tests are included


examples/getting_started/react/App.jsx
```diff
@@ -0,0 +1,93 @@
...
+      const offline = await client.networkStatus.isOffline();
+      // set network state
+      setIsOnline(!offline);
+    }
+    setInitialized(true);
+  }, []);
+
+  // set up network listener to 
+  // detect when app goes offline
+  client.networkStatus.onStatusChangeListener({
```
wtrocki: [Optional] It will be preferable to set that when creating client. 
We want to show here good patterns for using offix etc.

examples/getting_started/react/README.md
```diff
@@ -0,0 +1,20 @@
+# Getting started
```
wtrocki: [Question] Location wise. Wy we do not have simple examples/react?
Any reason to have getting_started folder?

examples/getting_started/react/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+```
+yarn install
+```
+
+## Setting up a server
+
+To get started, you will need to set up a GraphQL server.
```
wtrocki: So server will not be included? Without server, we will not be able to show conflicts

examples/getting_started/react/components/Todo/EditTodo.jsx
```diff
@@ -0,0 +1,66 @@
...
+  }
+
+  const handleUpdate = (e) => {
+    e.preventDefault();
+    
+    // get serialized inputs
+    const [title, description] = getInputs();
+
+    // execute mutation
+    editTodo({
```
wtrocki: [Question] Are we planning to add versioning here

wtrocki:
Example app is spotless. It works and runs as it should. I just have couple questions about how we going to proceed with that further.

examples/getting_started/react/components/Todo/Index.jsx
```diff
@@ -0,0 +1,27 @@
+import React, { useState } from 'react';
+import Content from './TodoContent';
```
darahayes: This is one of the potential drawbacks of default exports. We can basically assign arbitrary names to whatever was actually exported from the module. Sometimes, this makes things a little less obvious.

It might be a little bit clearer if we named this `TodoContent`

examples/getting_started/react/components/Todo/TodoContent.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+
+const Todo = ({ todo, editTodo, deleteTodo, toggleEdit }) => {
```
darahayes: Related to the previous comment: This filename here is `TodoContent` but the file is exporting a function called `Todo` but the parent component is also named `Todo` and that parent references this one as `Content`. While this is a small example, as an application grows this kind of approach to naming can quickly become confusing. This would be one of the arguments for totally avoiding using default exports but I'm not overly picky.

I guess this is not something you consciously did but you were probably just copy pasting stuff and it slipped past you. 😃 It would be a bit nicer if this were named `TodoContent`.

examples/getting_started/react/components/AddTodo.jsx
```diff
@@ -0,0 +1,63 @@
...
+
+const AddTodo = ({ addTodo, cancel }) => {
+
+  const titleRef = useRef();
+  const descriptionRef = useRef();
+
+  const handleSubmit = (e) => {
+    e.preventDefault();
+   
+    addTodo({
```
darahayes: Question: is there any strong reason that we pass in the `addTodo` mutation instead of creating/defining it in here? Defining it in here kinda encapsulates the entire creation logic within the component itself and it also has the potential benefit for the reader/user that they could be pointed to this file and see a really tidy, complete example of performing a mutation. What do you think?

examples/getting_started/react/README.md
```diff
@@ -0,0 +1,20 @@
+# Getting started
+
```
darahayes: It would be nice to have a one or two liner description of the app, what it does and what features/use cases of Offix it demonstrates.

examples/getting_started/react/README.md
```diff
@@ -0,0 +1,20 @@
...
+```
+yarn install
+```
+
+## Setting up a server
+
+To get started, you will need to set up a GraphQL server. 
+
+You can use the [Graphback runtime example](https://github.com/aerogear/graphback/tree/master/examples/runtime-example), to get
```
darahayes: Question: when I go to that example I don't see the tasks model, I see a user model down there. Are there additional steps needed to get it working?

wtrocki: For flexibility we might simply have model schema inside here and maybe use testx instead ?

examples/getting_started/react/components/AddTodo.jsx
wtrocki: I kinda liked that because it will be nice way to reference that from the docs - zero react stuff just pure offix logic, but no strong opinion

wtrocki: Also I see this pattern in couple production apps (and urigo react trainings)

darahayes: Fair enough, let's keep it as is :)

examples/getting_started/react/components/Todo/TodoContent.jsx
kingsleyzissou: This was actually one of those ones were I was debating the name for the component and what made the most sense. So at one stage it was Todo, then it became Content and then it became TodoContent and clearly I forgot to change it in each place. I think named exports will be slightly better, for the sake of maintenance.

examples/getting_started/react/components/AddTodo.jsx
kingsleyzissou: I agree about encapsulating all the creation logic. I was getting a memory leak error when I did that, because the addTodo creates a promise, if the app goes offline, the promise isn't resolved and when the component unmounts, React gives out a memory leak error. Maybe I could get your eyes on this to try and help debug

examples/getting_started/react/README.md
kingsleyzissou: Going to add a server that can be used by this example and React Native too

examples/getting_started/react/README.md
kingsleyzissou: I think the idea initially was to have a few smaller examples showcasing the different features. So there would have been a subscriptions folder etc. Since that doesn't seem to be the route we are going anymore, it will make more sense to do as you suggest and change the folder to `/examples/react`

kingsleyzissou:
@darahayes fixed the build issue and the duplicate versions of React. For now the examples folder is still part of the lerna build command. Not too sure what implications that will have on the publishing side of things though.

PR #353, opened 2020-02-04
fix: minor refactors to offix-cache

darahayes:
### Description

I originally intended to go as far as stopping cache updates being automatically applied but I decided to hold off on that for now. However, I had gotten started on doing some minor refactoring that does not change any functionality and I think it is worth merging just for readability/maintainability.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


wtrocki:
Having this in soo advanced state makes me think that there is very little needed for proposal to be implemented. This will definitely help to fix our api and docs. Still some issues will persist but that is ok considering we are doing architecture spikes

packages/offix-cache/src/createMutationOptions.ts
```diff
@@ -128,94 +133,101 @@ export const createMutationOptions = <T = {
...
+  } 
+  // this default catches the REFRESH case and returns an empty update function which does nothing
+  return () => {};
+};
+
+/**
+ * Generic cache update function that adds an item to a query that contains a list of items
+ * Might be exported in the future
+ */
+function addItemToQuery({ mutationName, updateQuery, idField = "id" }: CacheUpdateHelperOptions): MutationUpdaterFn {
```
wtrocki: Lets export this function

wtrocki:
Question wise. If we externalize cache updates. Are we going to always require people to use optimistic response helper as well? 

What you think about half measure of doing something like

 `offlineMutate({gql, variables, ...updateQueryWithOptimisticResponse()})

packages/offix-cache/src/createMutationOptions.ts
wtrocki: Lets rename it to `addResultToCache` or `addObjectToQuery`. Item sounds to specific

wtrocki:
There is assumption here. If we assume that mutation can only add or delete we can pass array or query variable and somehow implement proposal. Otherwise it will be messy to have it for multiple queries and then reusing them in global cache updates

packages/offix-cache/src/createMutationOptions.ts
darahayes: `addResultToQuery` sounds like a nice one

darahayes:
I really like the idea of the `addToCache` helper method that would give you back both `update` and the `optimisticResponse` options. This is the simplest way to remove things from the core but also keep things from being confusing. I would prefer to achieve it in a separate PR

packages/offix-scheduler/src/queue/OfflineQueue.ts
```diff
@@ -218,3 +217,14 @@ export class OfflineQueue<T> {
...
+
+/**
+ *
+ * Helper function to generate a simple alphanumeric ID for queue entries
+ */
+function ID() {
+  // Math.random should be unique because of its seeding algorithm.
+  // Convert it to base 36 (numbers + letters), and grab the first 9 characters
+  // after the decimal.
+  return "queue:" + Math.random().toString(36).substr(2, 9);
```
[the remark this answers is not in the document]
darahayes: Unnecessary IMO, this is a very simple approach to generating some IDs that will be unique enough for a client without adding any additional dependencies

packages/offix-cache/src/createMutationOptions.ts
```diff
@@ -128,94 +133,101 @@ export const createMutationOptions = <T = {
...
+      const operationData = data[mutationName];
+      try {
+        queryResult = cache.readQuery({ query, variables }) as any;
+      } catch (e) {
+        queryResult = {};
+      }
+
+      const result = queryResult[queryField];
+      if (result && operationData) {
+        // FIXME deduplication should happen on subscriptions
```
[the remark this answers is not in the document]
darahayes: I'll cover that in a subsequent PR 😃

PR #376, opened 2020-02-28
fix: Add mapping for conflict handling

wtrocki:
We found couple issues around the platform around mapping. 
This change introduces non breaking mapper that can be used in many places where variables are not defined on the root

### DIFF
--- a/packages/offix-cache/src/createOptimisticResponse.ts
+++ b/packages/offix-cache/src/createOptimisticResponse.ts
@@ -12,6 +12,7 @@ export interface OptimisticOptions {
   returnType: string;
   idField?: string;
   variables?: OperationVariables;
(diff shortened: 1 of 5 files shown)


packages/offix-cache/src/createOptimisticResponse.ts
```diff
@@ -49,9 +50,16 @@ export const createOptimisticResponse = (options: OptimisticOptions) => {
     __typename: "Mutation"
   };
 
+  let mappedVariables = variables;
+
+  if (options.inputMapper) {
```
darahayes: perhaps we could do the following?
```suggestion
  if (options.inputMapper && typeof a === 'function') {
```

Issue #367, opened 2020-02-25
Entities ids cache update workflow


lucasmez:
_This is not a bug report or a new feature idea, it is a question about cache update workflow for an offline app using Offix, so I apologize if I added this issue using the wrong label_

I am building an offline-first application that lets users switch between offline and online modes, and Offix looked like a very good solution for what I needed, until I ran into a problem: 

Users can create items in a nested list like so:

![Image of Yaktocat](http://maxdesign.com.au/maxdesign/wp-content/uploads/2015/04/folder-metaphor2.png) 

The GraphQL mutation I use to create items has the following signature: **CreateItem(itemTitle, parentId)** , parentId being the id of the item where the newly created item will be placed under. So, for example, to create the "sub-item 1" in the image above, I'd call **CreateItem("sub-item 1", <item 4 id>)**  Entities Ids must be generated by the database on the back-end instead of using client-generated ones.

I'm using Offix's OptimisticResponse helper, which generates an unique Id for each item using the  _client:UUID_ format.
If a users creates many nested items while offline, Offix will queue up many mutations with the second parameters (parentId) being populated with the Offix-generated ids. As soon as the user switches to online-mode, I assume those mutations will be sent to the server in parallel (with limitations imposed by the browser). But all of the parentIds for the mutations will still be the Offix-generated ids, which will result in the API returning an error for all of them, because entities with those parentIds do not exist in the database.

With that in mind, and assuming it is correct, here is the workflow I want to implement when a user creates items:

1. _User switches to Offline Mode._
2. _User creates many new items._
2. optimisticResponses  with client-generated ids are created for all of them.
3. Inside the CreateItem mutation update() function:
       `Add all optimistic responses to the cache using writeFragment() and writeQuery() so that components using useQuery() will be reactively updated`
4. _User can close/refresh the page many times. Offline data won't be lost since the optimistic responses were all cached in the persistent store._
4. _User switches to Online Mode._
5. Many mutations will be queued up by Offix. **Send them to the server asynchronously in series**. After each mutation call returns, and before sending the next one, do the following:
    `Replace all references to the client-generated ids from the call that just returned, with the real id returned from the server.`
    
      Example: Using the image above as reference, say there are 2 mutations queued up: 
      - _CreateItem("sub item 3", <real item 4 id>)_,  having a client-generated id of _client:subitem3_. 
      - _CreateItem("sub sub item 1", "client:subitem3")_, having a client-generated id of _client:subsub1_.

[...]

PR #381, opened 2020-03-03
fix: improve queue handling of optimistic ids + InputMapper

darahayes:
### Description

I've brought in a new dependency in order to solve these issues. [traverse](https://github.com/substack/js-traverse/blob/master/index.js) is a ~300 line dependency.

* `createOptimisticResponse` flattens `variables` this means no need for a mapper function. If this flattening does not work, then use `createOptimisticResponse` directly passing in the correct variables, or supply your own object.

* `replaceClientGeneratedIdsInQueue` is able to deep replace optimistic id values in queue entries no matter the location. This solves issues where `variables` isn't a flat object. (e.g. using `input` types). It also solves issues where entries in the queue reference other optimistically created objects. i.e. https://github.com/aerogear/offix/issues/367

Check unit tests for examples

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-cache/src/createMutationOptions.ts
+++ b/packages/offix-cache/src/createMutationOptions.ts
@@ -5,6 +5,11 @@ import { CacheUpdatesQuery } from "./api/CacheUpdates";
 import { getOperationFieldName, deconstructQuery } from "./utils/helperFunctions";
 import { isArray } from "util";
 
+export interface InputMapper {
+  deserialize: (object: any) => any;
+  serialize: (object: any) => any;
+}
+
 /**
  * Interface to overlay helper internals on top of mutation options.
  */
@@ -34,6 +39,14 @@ export interface MutationHelperOptions<T = {
    * For example for `modifyObject(value: String!): Object` value will be `Object`
    */
   returnType?: string;
+
+  /**
+   * [Modifier]
+   *
+   * Maps input objects for the cases if variables are not passed to the root
+   *
+   */
+  inputMapper?: InputMapper;
 }
 
 /**
@@ -88,7 +101,8 @@ export const createMutationOptions = <T = {
     returnType,
     operationType = CacheOperation.ADD,
     idField = "id",
-    context
+    context,
+    inputMapper
   } = options;
 
   if (returnType && !options.optimisticResponse) {
@@ -97,7 +111,8 @@ export const createMutationOptions = <T = {
       variables,
       returnType,
       operationType,
-      idField
+      idField,
+      inputMapper
     });
   }
 
--- a/packages/offix-client/package.json
+++ b/packages/offix-client/package.json
@@ -38,7 +38,7 @@
     "typescript": "3.7.5"
   },
   "dependencies": {
-    "@types/traverse": "^0.6.32",
+    "@types/traverse": "0.6.32",
     "apollo-cache-inmemory": "1.6.5",
     "apollo-cache-persist": "0.1.1",
     "apollo-client": "2.6.8",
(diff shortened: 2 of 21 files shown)


darahayes:
@wtrocki if you want to try it within the datasync-starter, check the `remove-offix-hacks` branch https://github.com/aerogear/datasync-starter/tree/remove-offix-hacks

wtrocki:
I kinda think that current PR is ok but we simply need abstraction that will be able to do flatten or anything else to deserialize and deserialize data. There is risk of having some invalid data send to server or failing that will be hard for people to debug and they will log issues. We are not adding anything to docs so this entire behavior will be really hard to understand.

If we keep abstraction we can use no abstraction by default and flatten one for the showcase. This will be also easier to document - config option etc. 
This is better hack than embedding flattening into create optimistic response.
We cannot avoid using flatten in current scenario and this could cause some issues that we cannot predict.

wtrocki:
Verified on top of datasync-starter. Applied suggestion

wtrocki:
Went through the integration tests. The fact that they are failing with this change is worrying. We should really apply this approach as strategy rather than having it in the code.

packages/offix-cache/src/createMutationOptions.ts
```diff
@@ -5,6 +5,11 @@ import { CacheUpdatesQuery } from "./api/CacheUpdates";
 import { getOperationFieldName, deconstructQuery } from "./utils/helperFunctions";
 import { isArray } from "util";
 
+export interface InputMapper {
```
wtrocki: Let's add some docs. for better api documentation

packages/offix-cache/src/createMutationOptions.ts
```diff
@@ -34,6 +39,14 @@ export interface MutationHelperOptions<T = {
    * For example for `modifyObject(value: String!): Object` value will be `Object`
    */
   returnType?: string;
+
+  /**
+   * [Modifier]
+   *
+   * Maps input objects for the cases if variables are not passed to the root
```
wtrocki: It is hard to understand it. I think an example will work better than anything else.

packages/offix-client/package.json
```diff
@@ -38,6 +38,7 @@
     "typescript": "3.7.5"
   },
   "dependencies": {
+    "@types/traverse": "^0.6.32",
```
wtrocki:
```suggestion
    "@types/traverse": "0.6.32",
```

packages/offix-client/package.json
```diff
@@ -48,7 +49,8 @@
     "offix-cache": "0.13.2",
     "offix-conflicts-client": "0.13.2",
     "offix-offline": "0.13.2",
-    "offix-scheduler": "0.13.2"
+    "offix-scheduler": "0.13.2",
+    "traverse": "^0.6.6"
```
wtrocki:
```suggestion
    "traverse": "0.6.6"
```

packages/offix-cache/src/createMutationOptions.ts
wtrocki: We can also mention that there is traverse mapper that will flatten your input types

wtrocki:
I have use case how where I need to create objects like this:

```graphql
type ParcelShop  @auth(type: 'mutation') {
  id: String
  shopType: String
  name: String
  keyword: String
  closurePeriods: [String]
  openingTimes: [OpeningTimes]
  geoLocation: GeoLocation 
  address: Address
}
```

This objects have multiple relationships. 
How flatten will behave in this case - it flattens everything or just first node?

PR #387, opened 2020-03-06
fix: react example app build issue

kingsleyzissou:
### Description
The example app build was breaking on my end yesterday too, so I added the line back into the package.json that seems to fix things.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [ ] tests are included


wtrocki:
Triggered build again as it passes locally

PR #388, opened 2020-03-06
Bundle size

kingsleyzissou:
### Description

- Add CI stage for bundle sizes
- Check can be run from root level and from package level
- at pacakage level, `size:why` script is available to view package dependencies (see screenshot)

image

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [ ] tests are included


wtrocki:
I can see why persist is taking so much. It is very small library so it might be possible that it incorporates some extra packages from Apollo that are not needed. On the other hand we can also generate esm of our stuff as this will put it's size by half. 

I think this will look much simpler if we create single dist over the apollo package and dist with apollo package and then compare it to the apollo package itself. Our packages are in non minified form so it is artificaly large

kingsleyzissou:
@wtrocki the package that we use minifies and gzips our packages. The screenshot there is just from all the dependencies in the offix-client package.

image

This is from bundlephobia that shows something similar to the initial screenshot:

image

wtrocki:
That is really good! I would call it small in my standards

kingsleyzissou:
@darahayes I'll do that now quick

PR #393, opened 2020-03-09
feat: allow for adding conflict listeners after client creation

darahayes:
### Description

Adds a `CompositeConflictListener` class that always gets passed down to the `ConflictLink`. This allows us to add/remove listeners as needed. It's quick and dirty but it works and it has an integration test.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-client/src/apollo/conflicts/CompositeConflictListener.ts
+++ b/packages/offix-client/src/apollo/conflicts/CompositeConflictListener.ts
@@ -1,0 +1,38 @@
+import { ConflictListener, ConflictResolutionData } from "offix-conflicts-client";
+
+
+/**
+ * Composite Conflict Listener class that can accept register and remove individual listener functions as needed
+ * Gets passed down to the conflict link
+ */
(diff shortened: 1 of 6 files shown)


packages/offix-client/src/apollo/conflicts/CompositeConflictListener.ts
```diff
@@ -0,0 +1,38 @@
+import { ConflictListener, ConflictResolutionData } from "offix-conflicts-client";
+
+
+/**
+ * Composite Conflict Listener class that can accept register and remove individual listener functions as needed
+ * Gets passed down to the conflict link
+ */
+export class CompositeConflictListener implements ConflictListener {
```
kingsleyzissou: Awesome, does this mean the changes aren't even breaking changes?

darahayes: exactly, you can still pass your original `conflictListener` in options if desired but it'll get registered into this Composite one.

PR #402, opened 2020-03-20
Runtime config validation Closes #283

### LINKED ISSUE #283: Validate config at runtime
## Feature Request

Idea is to use runtime interface checker for js code

https://github.com/gristlabs/ts-interface-checker

**Is your feature request related to a problem? Please describe.**
Please describe the problem you are trying to solve.

**Describe the solution you'd like**
Please describe the desired behavior.

**Describe alternatives you've considered**
Please describe alternative solutions or features you have considered.

Eunovo:
### Description
Work in progress pull request for runtime config validation as specified [here](https://github.com/aerogear/offix/issues/283)

No tests have been written yet.


packages/offix-client/src/config/ApolloOfflineClientConfig.ts
```diff
@@ -17,11 +17,29 @@ import { CacheUpdates } from "offix-cache";
...
+  offlineQueueListener?: ApolloOfflineQueueListener;
+  conflictStrategy: ConflictResolutionStrategy;
+  conflictProvider: ObjectState;
+  networkStatus?: NetworkStatus;
+  terminatingLink: ApolloLink | undefined;
+  cacheStorage: PersistentStore;
+  offlineStorage: PersistentStore;
+  conflictListener: CompositeConflictListener;
+  mutationCacheUpdates?: CacheUpdates;
+  cachePersistor?: CachePersistor;
```
kingsleyzissou: I think this is what is breaking the build,
it should be:
```
...
cacheStorage: PersistentStore<PersistedData>,
offlineStorage: PersistentStore<PersistedData>,
...
cachePersistor?: CachePersistor<object>
```

However, at that point, it may just be better to use the original `ApolloOfflineClientOptions` interface as this interface would then be identical.

packages/offix-client/src/config/ApolloOfflineClientConfig-ti.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * This module was automatically generated by `ts-interface-builder`
+ */
+import * as t from "ts-interface-checker";
+// tslint:disable:object-literal-key-quotes
+
+export const ApolloOfflineClientConfigIntf = t.iface(["ApolloOfflineClientOptions"], {
```
kingsleyzissou: I tried this using the react example app, and it complains about `ApolloOfflineClientOptions` being unknown. So it seems that the interface checker might not understand anything beyond basic types. This means we might need an interface checker for all of the classes in this file.
i.e. we'd have to create an `ApolloOfflineClientOptions-ti` file, the same for an `ApolloOfflineQueueListener-ti` file and so on.

cc @darahayes and @wtrocki what are your thoughts on going down the rabbit hole of creating an interface checker for each of these classes?

packages/offix-client/src/config/ApolloOfflineClientConfig.ts
kingsleyzissou: I see that ts-interface-builder doesn't support generics though, so I see now why there was a need to create a new interface.

Eunovo: Exactly! `ApolloOfflineClientConfigIntf` also fails because of its fields that are generic types

wtrocki:
My recommendation will be to update this to use a couple of ifs for the most important parameters

packages/offix-client/src/config/ApolloOfflineClientConfigValidator.ts
```diff
@@ -0,0 +1,17 @@
+import { ApolloOfflineClientConfig } from "./ApolloOfflineClientConfig";
+import { ConfigError } from "./ConfigError";
+
+export default function(config: ApolloOfflineClientConfig) {
+    checkFields(config, [
+        "conflictStrategy", "conflictProvider", "cacheStorage",
```
kingsleyzissou: `conflictStrategy` and `conflictProvider` have fallbacks set when the client Conflig is created, 
so I think those two can be removed. 

`cacheStorage` and `offlineStorage` are also created before this, if nothing was defined, but I think it is still worthwhile keeping those in there. The nice thing about how you have done things is that it is dynamic, so we can add or remove things easily if we need to.

packages/offix-client/src/config/ApolloOfflineClientConfigValidator.ts
```diff
@@ -0,0 +1,17 @@
...
+        "offlineStorage", "conflictListener", "cache"
+    ]);
+}
+
+function checkFields(obj: any, fields: string[]) {
+    fields.forEach((fieldName) => {
+        if (obj[fieldName] === undefined) {
+            throw new ConfigError(`${fieldName} is required`, fieldName);
```
kingsleyzissou: This is just nitpicking, but can we get a slightly more meaningful message here, something like:
`Offix client config error: ${fieldName} is required`

packages/offix-client/src/config/ApolloOfflineClientConfigValidator.ts
Eunovo: Thank you. That's what I was going for. I tried to generate the list of required fields at runtime from the given type, i.e `ApolloOfflineClientConfig`, to no avail. So, I settled for this instead.

packages/offix-client/src/config/ApolloOfflineClientConfigValidator.ts
```diff
@@ -0,0 +1,19 @@
+import { ApolloOfflineClientConfig } from "./ApolloOfflineClientConfig";
+import { ConfigError } from "./ConfigError";
+
+export default function(config: ApolloOfflineClientConfig) {
```
darahayes: nit: we tend to avoid default exports in Offix. Additionally, anonymous functions (functions with no name) are not great because if an error is thrown from them the stack trace will say something like `anonymous` function instead of the name of the function.

Could we change it to something like `export function validateConfig(config: ApolloOfflineClientConfig)`?

Eunovo: I'll change it

PR #435, opened 2020-04-24
fix: id overwrite issue in optimistic responses

### LINKED ISSUE #430: createOptimisticResponse CacheOperation.ADD overwrites id prop
createOptimisticResponse with CacheOperation.ADD creates its own id using generateClientId. If they id prop already has a value can we not blow the existing value away? I already generated a id for this new element and I want to keep it in the format I provided. I know I can work around this using the default apollo update function, this would be a nice thing to have.

Thanks
Mitch

darahayes:
### Description

Fixes #430 

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


wtrocki:
Verified in react app

PR #484, opened 2020-05-28
fix: remove precommit

wtrocki:
Trying to make any change in offix requires tests to pass. 
This is very very bad for general community that cannot create PR without failing tests. I know that works nice for maintainers but we should be focusing on community more than just ourselves.

See discussion on offix chat for more context


kingsleyzissou:
Sure we can set it up locally for ourselves

PR #485, opened 2020-05-28
Added debug logging to offline store

shrey:
### Description

Added logging to the offline store using debug library

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-scheduler/src/store/OfflineStore.ts
+++ b/packages/offix-scheduler/src/store/OfflineStore.ts
@@ -1,7 +1,8 @@
 import { PersistentStore, PersistedData } from "./PersistentStore";
 import { QueueEntryOperation, QueueEntry } from "../queue";
 import { OfflineStoreSerializer, DefaultOfflineSerializer } from "./OfflineStoreSerializer";
-
+import Debug from "debug";
+const debug = Debug("Store: ");
 /**
  * Abstract Offline storage
  */
@@ -35,14 +36,20 @@ export class OfflineStore<T> {
    * @param entry - the entry to be saved
    */
   public async saveEntry(entry: QueueEntryOperation<T>) {
-    const serialized = this.serializer.serializeForStorage(entry);
-    const offlineKey = this.getOfflineKey(entry.qid);
-    // only add the offline key to the arrray if it's not already there
-    if (!this.arrayOfKeys.includes(offlineKey)) {
-      this.arrayOfKeys.push(offlineKey);
-      await this.storage.setItem(this.offlineMetaKey, this.arrayOfKeys);
-    }
-    await this.storage.setItem(offlineKey, serialized);
+    try{
+      const serialized = this.serializer.serializeForStorage(entry);
+      const offlineKey = this.getOfflineKey(entry.qid);
+      // only add the offline key to the arrray if it's not already there
+      if (!this.arrayOfKeys.includes(offlineKey)) {
+        this.arrayOfKeys.push(offlineKey);
+        await this.storage.setItem(this.offlineMetaKey, this.arrayOfKeys);
+      }
+      await this.storage.setItem(offlineKey, serialized);
+      }catch(error){
+        debug("Error Occured while Saving");
+        return;
+      }
+    debug("Saved Item");
   }
 
   /**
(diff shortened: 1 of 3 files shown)


shrey:
@wtrocki I wonder if it's throwing an error in my try and catch blocks

packages/offix-scheduler/src/store/OfflineStore.ts
```diff
@@ -1,7 +1,8 @@
 import { PersistentStore, PersistedData } from "./PersistentStore";
 import { QueueEntryOperation, QueueEntry } from "../queue";
 import { OfflineStoreSerializer, DefaultOfflineSerializer } from "./OfflineStoreSerializer";
-
+import Debug from "debug";
+const debug = Debug("Store: ");
```
wtrocki, line 5: I would go with the wrapper DataSync:Store: etc. Space after makes this unusable.

packages/offix-scheduler/src/store/OfflineStore.ts
```diff
@@ -1,7 +1,8 @@
 import { PersistentStore, PersistedData } from "./PersistentStore";
 import { QueueEntryOperation, QueueEntry } from "../queue";
 import { OfflineStoreSerializer, DefaultOfflineSerializer } from "./OfflineStoreSerializer";
-
+import Debug from "debug";
```
wtrocki, line 4: It is small letter

packages/offix-scheduler/src/store/OfflineStore.ts
```diff
@@ -35,50 +36,57 @@ export class OfflineStore<T> {
...
+    
+      const serialized = this.serializer.serializeForStorage(entry);
+      const offlineKey = this.getOfflineKey(entry.qid);
+      // only add the offline key to the arrray if it's not already there
+      if (!this.arrayOfKeys.includes(offlineKey)) {
+        this.arrayOfKeys.push(offlineKey);
+        await this.storage.setItem(this.offlineMetaKey, this.arrayOfKeys);
+      
+      await this.storage.setItem(offlineKey, serialized);
+      debug("Saved Item");
```
wtrocki, line 48: Trick is to actually serialize item :D

packages/offix-scheduler/src/store/OfflineStore.ts
```diff
@@ -35,50 +36,57 @@ export class OfflineStore<T> {
...
-    this.storage.setItem(this.offlineMetaKey, this.arrayOfKeys);
-    const offlineKey = this.getOfflineKey(entry.qid);
-    await this.storage.removeItem(offlineKey);
+    
+        this.arrayOfKeys.splice(this.arrayOfKeys.indexOf(entry.qid), 1);
+        this.storage.setItem(this.offlineMetaKey, this.arrayOfKeys);
+        const offlineKey = this.getOfflineKey(entry.qid);
+        await this.storage.removeItem(offlineKey);
+    
+    debug("Removed Item");
```
wtrocki, line 63: Same here

wtrocki:
Not sure why this was closed. I get that you would like to create new one?

wtrocki:
I think to resolve errors you need to execute yarn lint-fix to resolve trailing whitespace (your editor will format that properly as well)

packages/offix-scheduler/src/store/OfflineStore.ts
shrey: I saw it on a stack overflow post [https://stackoverflow.com/questions/47724165/how-to-use-the-debug-module-with-typescript](post), but I'll try it with small letter too

wtrocki: Stack overflow :)

shrey: and the yarn lint-fix is not running, should I just externally remove that space?

shrey: @wtrocki

wtrocki: There are other things to fix. spaces are least concern really. VScode eslint will help as well

kingsleyzissou: You could change the import to:

`import debug from 'debug'`

And then call the variable something like `info` or `log`. I prefer info, feels more descriptive to me.

So:

`const info = debug(...)`

packages/offix-scheduler/src/store/OfflineStore.ts
kingsleyzissou: We could go with `Datasync:Store`, but that will be the flag we have to use to make this information visible. i.e. we would have to run `DEBUG=DataSync:Store yarn start` or whatever script you are using.

Maybe we could shorten it and keep it descriptive still.

Something like:
`const info = debug('offix:store')` or const info = debug('sync:store')`

packages/offix-scheduler/src/store/OfflineStore.ts
kingsleyzissou: You can add items in as added arguments. So,
`debug("Saved item", serialized)` - note if we change the name of the logger function, use that
name instead of `debug`.

kingsleyzissou:
@shrey I've added some feedback here to help a little with this. As @wtrocki mentioned, we can sort the linting out as a last step.

kingsleyzissou:
@shrey since this has been sitting here for a while, I am going to close the PR.

If you still want to implement the suggestions we've made, feel free to re-open the pull request.

PR #495, opened 2020-06-11
Offix.Next()

Eunovo:
### Description

This PR contains work on offix-datastore

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


wtrocki:
Amazing. Let's keep this here so we wil track progress and commits

kingsleyzissou:
Wow looks like some amazing work

packages/offix-datastore/tests/mock.graphql
```diff
@@ -0,0 +1,21 @@
+""" @model """
+type Note {
+    id: ID!
+    title: String!
+    description: String!
+    """
+    @oneToMany field: 'noteComment', key: 'noteId'
```
kingsleyzissou, line 7: Should this be `key: 'commentId'`, rather than `noteId` or whatever the equivalent is in the db?

Eunovo: I just copied this file from graphback to use for testing. Do we need to focus on the `@oneToMany` annotation? I'm not using it. At least not now.

kingsleyzissou: No don't stress about it. I'm just going through the code and started with the models to get an idea of what is going on.

Looks like you have done an incredible amount of work.

Were you able to solve the db concurrency?

packages/offix-datastore/tests/DataStore.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    configure(`${__dirname}/mock.graphql`);
+});
+
+afterEach(async () => {
+    await new Promise((resolve, reject) => {
+        indexedDB.deleteDatabase("offix-datastore").onsuccess = event => resolve();
+    });
+});
+
+test("Setup client db with provided schema", async () => {
```
kingsleyzissou, line 40: @Eunovo This test is failing for me locally. It looks like the test is timing out because the indexedDB.deleteDatabase function in the afterEach cycle is getting blocked and never resolving the promise.

I'm not sure if it will be the same in circleci becaues the pipeline failed before it got to the tests.

Do you mind fixing the lint issues so we can see if the tests run okay on the pipeline.

Are the tests all passing for you?

kingsleyzissou: I think the same is happening for the second test too

packages/offix-datastore/tests/mock.graphql
Eunovo: Yes. The problem came from db connections I opened directly in the tests, these connections didn't have an `onversionchange` handler so they had to be closed manually or they block any upgrade requests

packages/offix-datastore/tests/DataStore.test.ts
Eunovo: All right

packages/offix-datastore/src/storage/IndexedDBStorage.ts
```diff
@@ -0,0 +1,81 @@
...
+        const storeName = getStoreNameFromModelName(modelName);
+        return db.transaction(storeName, "readwrite")
+            .objectStore(storeName);
+    }
+
+    async save(model: Model) {
+        const store = await this.getStore(model.__typename);
+        const persistedModel = { ...model, id: generateId() };
+        const key = await this.convertToPromise<IDBValidKey>(store.add(persistedModel));
+        const result = await this.convertToPromise<PersistedModel>(store.get(key));
```
kingsleyzissou, line 58:
```suggestion
        return this.convertToPromise<PersistedModel>(store.get(key));
```

kingsleyzissou:
@Eunovo, this is some very impressive stuff. I am using the tests as an entry point to understand the code.

But looks like you have done a lot of work in a very short time.

packages/offix-datastore/tests/DataStore.test.ts
Eunovo: The first test in `DataStore.test.ts` times out for me

kingsleyzissou:
@Eunovo yeah it looks like the same test is failing in circle ci.

Also, it tooks like there may be an issue with the build config, it looks like the tests are being compiled too (they're in the dist folder).

Eunovo:
> @Eunovo yeah it looks like the same test is failing in circle ci.
> 
> Also, it tooks like there may be an issue with the build config, it looks like the tests are being compiled too (they're in the dist folder).

Yeah, I added the included the tests in tsconfig.json so that I could run `yarn lint-fix`. I have removed it

Eunovo:
@kingsleyzissou  The same test fails in circle CI. I have actually been ignoring it on my end. I'm not sure about the cause of this "time out". I added a log to indicate the conclusion of this test and it fired. The `afterEach` fixture is used for other tests too and they don't time out.

wtrocki:
@Eunovo I went thru the code and I'm really really impressed about what you have done in such short timeframe. Really good job!

Really looking forward for the chat next week. Let me know if you would like to have it later/earlier depending on the progress.

packages/offix-datastore/src/DataStore.ts
```diff
@@ -0,0 +1,39 @@
+import { buildSchema } from "graphql";
+import { readFileSync } from "fs";
+import { createDefaultStorage, Storage } from "./storage";
+import { extractModelsFromSchema, Model, PersistedModel } from "./models";
+import { createPredicate } from "./predicates";
+
+let storage: Storage;
+
+export function configure(schemaLocation: string, schemaVersion: number = 1) {
+    const schemaText = readFileSync(schemaLocation, "utf8");
```
wtrocki, line 10: I love that this is schema aware. 
So all model information will be based on the schema.

Would this happen on runtime or you see us having CLI that will store some extra metadata for models?

packages/offix-datastore/src/DataStore.ts
```diff
@@ -0,0 +1,39 @@
...
+    const schema = buildSchema(schemaText);
+    const models = extractModelsFromSchema(schema);
+    storage = createDefaultStorage(models, schemaVersion);
+}
+
+export function save(model: Model): Promise<PersistedModel> {
+    return storage.save(model);
+}
+
+export function query(model: Model, predicateFunction?: Function) {
```
wtrocki, line 20: So in this approach we tradig off ability to issue GraphQL Queries towards our own predicate function? Would you see ability to retrofit GraphQL queries or this is something completely different.

packages/offix-datastore/src/DataStore.ts
```diff
@@ -0,0 +1,39 @@
...
+    return storage.update(model);
+}
+
+// TODO delete all?
+export function remove(model: PersistedModel, predicateFunction?: Function) {
+    if (!predicateFunction) {return storage.remove(model);}
+
+    const modelPredicate = createPredicate(model);
+    const predicate = predicateFunction(modelPredicate);
+    return storage.remove(model, predicate);
```
wtrocki, line 38: On the backend, we use approach to set `_deleted` flag to false. This gives numerous benefits for the conflict resolution

packages/offix-datastore/src/models.ts
```diff
@@ -0,0 +1,21 @@
...
+    __typename: string;
+}
+
+export interface PersistedModel extends Model {
+    id: string;
+}
+
+export function extractModelsFromSchema(schema: GraphQLSchema): Model[] {
+    const types = getUserTypesFromSchema(schema);
+    const userTypes = types.filter((modelType: GraphQLObjectType) => parseMarker("model", modelType.description));
```
wtrocki, line 15: Nice! I like how you can enable model processing. Cannot comprehend how many good stuff this have. I think this comes back as very strongly tied to graphback. To not tie it I think we can:

- In the graphback client we can generate models file for offix datastore. 
- For offix we going to document how to create models files by themselves so you do not need schema and very performance intensive operations happening on the client - it will be simple JSON

WDYT?

packages/offix-datastore/src/predicates/PredicateFunctions.ts
```diff
@@ -0,0 +1,62 @@
+export abstract class PredicateFunction {
+    public filter(models: any[]) {
+        return models.filter((m) => this.evaluate(m));
+    }
+
+    public abstract evaluate(model: any): boolean;
+}
+
+export class ModelFieldPredicate extends PredicateFunction {
```
wtrocki, line 9: Absolutely love this. That is what we support on the server. 
https://graphqlcrud.org/docs/next/find

packages/offix-datastore/src/predicates/PredicateFunctions.ts
```diff
@@ -0,0 +1,62 @@
...
+    public evaluate(model: any) {
+        return this.operator(model[this.key], this.value);
+    }
+}
+
+export interface ExpressionOperator {
+    operate(previousResult: boolean, currentResult: boolean): boolean;
+}
+
+export const ExpressionOperators = {
```
wtrocki, line 27: Total rocket science! I'm not longer impressed this is so clever that I cannot comprehend it :D

packages/offix-datastore/src/DataStore.ts
Eunovo: I'm not sure about how much information from the schema we will need yet, right now, I only extract model types and create object stores.
Right now, I'm leaning generating types for models with codegen cli. But for most of the operations so far only the model name is required `__typename` which is used to determine the store so we might just use model names in the API. What I have here is a rough sketch to that I hope we can work on.

packages/offix-datastore/src/storage/IndexedDBStorage.ts
```diff
@@ -0,0 +1,102 @@
...
+            openreq.onsuccess = () => {
+                const db = openreq.result;
+                db.onversionchange = function() {
+                    this.close();
+                    // alert("Please reload the page.");
+                };
+                resolve(db);
+            };
+
+            openreq.onupgradeneeded = () => {
```
wtrocki, line 24: Speechless! You have even done support for the store upgrades! 
WOW. Just wow. I cannot wait to test this!

packages/offix-datastore/src/DataStore.ts
Eunovo: Makes sense to use this. But I have not implemented synchronisation and conflict resolution yet.

wtrocki:
@Eunovo Do you use any sample app for this? How it is possible to try this out. 
I see the tests and they getting a lot of contexts but nothing beats the live app.

packages/offix-datastore/src/DataStore.ts
Eunovo: AWS uses something similar. Currently, it is not possible to run graphql queries against the DataStore. Although, this is something we can add.

packages/offix-datastore/src/DataStore.ts
wtrocki: From my side this is perfect. You will need to understand model names and their relationships along with the ID's for effective predicates. 

All off that can be easily done after so no worries.

packages/offix-datastore/src/models.ts
Eunovo: Yes! This is great. I was thinking codegen before but using graphback client is better. Another reason I check schema though is to get the object stores to create and delete. The user could provide a list of models instead.

Eunovo:
Okay, @wtrocki I was going to set up sample app after doing synchronization but if you want one, I set up an app for next week.

packages/offix-datastore/src/predicates/PredicateFunctions.ts
Eunovo: Thank you! I just avoided switch statements. Doing it this way all us to obey the Open-Closed principle. Now, you can add an `ExpressionOperator` without modifying the `PredicateExpression` class

packages/offix-datastore/src/storage/IndexedDBStorage.ts
Eunovo: Thank you! it just occurred to me. I comment out the alert because it causes errors during tests. There should be a way to resolve this though, I'm just putting off for now

packages/offix-datastore/src/storage/adapters/IndexedDBStorage.ts
```diff
@@ -1,11 +1,11 @@
...
-import { PredicateFunction } from "../predicates";
+import { Model, PersistedModel } from "../../models";
+import { IStorageAdapter } from "../Storage";
+import { getStoreNameFromModelName } from "../core";
+import { PredicateFunction } from "../../predicates";
 
 const DB_NAME = "offix-datastore";
 
-export class IndexedDBStorage implements Storage {
+export class IndexedDBStorage implements IStorageAdapter {
```
kingsleyzissou, line 8: Very cool, this makes it easier to add in async storage for React Native at a later stage.

packages/offix-datastore/src/utils/PushStream.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export class ObservablePushStream<T> implements PushStream<T> {
+    private observable: Observable<T>;
+    private observer: any;
+
+    constructor() {
+        this.observable = new Observable(observer => {
+            this.observer = observer;
+        });
+        // disable @typescript-eslint/no-empty-function
```
kingsleyzissou, line 20:
```suggestion
        // eslint-disable-next-line @typescript-eslint/no-empty-function
```

docs/ref-datastore-getting-started.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Using the DataStore inside your application
+
+To work with DataStore, you need to pass your models to DataStore and allow it setup stores for each model.
+
+```typescript
+import { configure } from 'offix-datastore';
+
+configure([
+    { __typename: "Task" }
```
wtrocki, line 24: We need new schema format. models: [..]

Eunovo:
@kingsleyzissou Please force push your work to next again. I wasn't aware you rebased and I overwrote your work.

kingsleyzissou:
Don’t stress I only rebased it on top of master. Didn’t make any changes

wtrocki:
This branch should be static - no active work should be happening here. We can merge other branch and find best time to rebase - when no one has changes

wtrocki:
WE merged this changes in separate PR.

Issue #499, opened 2020-06-15
Offix 2.0 top level goals


wtrocki:
## Top level requirements

1. Ability to work with the local data without interacting with the cache/building GraphQL queries

1. Ability to see offline changes and being able to differentiate between change that was made offline

3. Ability to work without building Optimistic responses or relying on client cache.

4. Simplicity to work with subscriptions on the root level and populate the data back to passive views.

5. Ability to work in a simplified data interface that is utilizing https://graphqlcrud.org capabilities

6. Support any types of data containers like Pagination container, DataSync container etc. 

7. First-class support for relationships 

## Notes

This is just created as proposal any edits/suggestions welcomed

wtrocki:
Since we already have datastore prototype let's discuss those and see if we can build some early set of issues to build this.

Eunovo:
> Ability to see offline changes and being able to differentiate between change that was made offline

@wtrocki from the DataStore perspective, changes that will be considered offline are those that have not be synced with the server. We can have API that tells if a particular change has been synced. Like
```ts
const item = (await DataStore.query(Note))[0];
console.log(item.isSynced)
```
We can implement this by checking if the `lastUpdated` is before our last sync with the server.
WDYT?

wtrocki:
@Eunovo Exactly this! There is no way you can do that now in the offix 1.0 as we are dealing with the cache and optimistic response flag is not being returned from client (as everything is the graphql query driven)

kingsleyzissou:
Would nr 2 get rid off the need for an offline queue, in the sense that we would could have a computed list of changes based on the items that have not yet been synced?

Or would there still be value in keeping an offline queue?

kingsleyzissou:
What I would like to see is a simple setup with little fuss over how to configure the client and the store and all of that.

It would be nice if we could have a simple and clean way to setup the offix client, that would be amazing.

wtrocki:
> Or would there still be value in keeping an offline queue?

We need offline queue to stage original changes in history as datastore entities are in merged form. Also it is needed for rollback etc.

wtrocki:
> It would be nice if we could have a simple and clean way to setup the offix client, that would be amazing.

Yes. Single is package and flags for feature enablement. Custom storage implementation with default available out of the box

PR #504, opened 2020-06-15
Pagination container cache updates.

wtrocki:
Add ability to provide cache updates for the pagination/deltasync

### DIFF
--- a/packages/offix-cache/src/createMutationOptions.ts
+++ b/packages/offix-cache/src/createMutationOptions.ts
@@ -75,7 +75,7 @@ interface CacheUpdateHelperOptions {
  * Set of parameters used to generate the update function to
  * update the cache for a given operation and query.
  */
-interface CacheUpdateOptions extends CacheUpdateHelperOptions{
+interface CacheUpdateOptions extends CacheUpdateHelperOptions {
   /**
    * Defines operation type used to make appropriate changes in cache
    *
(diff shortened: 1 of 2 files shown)


wtrocki:
@kingsleyzissou As we talked about it this is not going to land anywhere - it is very hacky change and it will work only with the https://graphqlcrud.org spec (like entire offix really)

The way I see this being accepted would be to make sane decision to deprecate cache package for general use and make it work for graphback stuff.

packages/offix-cache/src/createMutationOptions.ts
```diff
@@ -190,11 +190,22 @@ function addItemToQuery({ mutationName, updateQuery, idField = "id" }: CacheUpda
...
-        }
+          if(result.find){
+              const foundItem = !result.find(function (item) {
+                  return item[idField] === operationData_1[idField];
+              });
+              if (foundItem) {
+                  result.push(operationData_1);
+              }
+
+          } else if(result.items && result.items.find){
```
wtrocki, line 201: assumes `items` exsit for Pagination and sync (we use the same name)

wtrocki:
CC @Eunovo for context. This is why we want datastore. It is not easy to work wit the cache. Updates are unpredictable and codebase is really hard to manage.


This fix is required to have https://github.com/aerogear/offix/issues/499 working

packages/offix-cache/src/createMutationOptions.ts
kingsleyzissou: What is the impact if pagination and sync is not wanted?

kingsleyzissou:
> The way I see this being accepted would be to make sane decision to deprecate cache package for general use and make it work for graphback stuff.

I guess it that kind of makes sense because the cache helpers only seem to work with our stack because we make assumptions on how the inputs and return types are formatted (according to our stack). So custom update functions would be required when using a different backend anyway.

packages/offix-cache/src/createMutationOptions.ts
wtrocki: Zero. This will just make it work (so this else will never appear).
This can be also used upstream. if you want to wrap your offline data into some container you just need to name variable `items`.

Let's make build pass and release it.

wtrocki: Actually there is problem here. If someone will have named field `find` it will stop working for them. Chance for that is very low though.

I had much simpler implementation before that used `items` for checking but that can actually be field on the object.

wtrocki:
@kingsleyzissou Ready for review. 

Verifcation would be to try sample app with the compiled version. I have tested it on the datasync-starter with the container and without container (master)

wtrocki:
Added small fix for 490

wtrocki:
Released 0.16.0-dev1

PR #518, opened 2020-06-19
fix: allow prereleases without bumps

wtrocki:


wtrocki:
Replicating changes from graphback that were tested. No need to review

PR #524, opened 2020-06-23
Cache pagination

kingsleyzissou:
### Description
- Updated mutation and subscription cache helpers to provide support for field level cache updates.
- Updated documents to reflect new changes.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [ ] tests are included

### DIFF
--- a/docs/ref-cache.md
+++ b/docs/ref-cache.md
@@ -76,9 +76,28 @@ const options = createMutationOptions(mutationOptions);
 apolloClient.mutate(options);
 ```
 
-> NOTE: Cache helpers currently support only GraphQL Queries that return arrays.
-> For example `getUsers():[User]`.
-> When working with single objects returned by Queries we usually do not need use any helper as Query will be updated automatically on every update
+### Pagination and relationships
+
+It may be necessary to update a field within the query and not the entire query, for example with relationships. In order to do this, it is necessary to provide the name of the field being updated with the `returnField` parameter.
+
+```javascript
+const mutationOptions = {
+  mutation: ADD_COMMENT,
+  variables: {
+    title: 'comment title'
+  },
+  updateQuery: {
+    query: GET_TASK,
+    variables: {
+      filterBy: 'some filter'
+    }
+  },
+  returnType: 'Comment',
+  operationType: CacheOperation.ADD,
+  idField: 'id',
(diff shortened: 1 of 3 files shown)


docs/ref-cache.md
```diff
@@ -76,9 +76,28 @@ const options = createMutationOptions(mutationOptions);
 apolloClient.mutate(options);
 ```
 
-> NOTE: Cache helpers currently support only GraphQL Queries that return arrays.
-> For example `getUsers():[User]`.
-> When working with single objects returned by Queries we usually do not need use any helper as Query will be updated automatically on every update
+### Pagination and relationships
+
+It may be necessary to update a field within the query and not the entire query, for example with relationships. In order to do this, it is necessary to provide the name of the field being updated with the `returnField` parameter.
```
wtrocki, line 81:
```suggestion

Offix by default assumes that object returned by mutations will contain just your data. In situations when you need to wrap your data into some container to provide pagination and other information offix will require an additional parameter. For example when API returns TaskPage etc. we actual data might be returned over `items`.

When this query will be cached it may be necessary to update a field within the query and not the entire query, for example with relationships. In order to do this, it is necessary to provide the name of the field being updated with the `returnField` parameter.
```

packages/offix-cache/src/createMutationOptions.ts
```diff
@@ -242,18 +250,21 @@ function deleteItemFromQuery({ mutationName, updateQuery, idField = "id" }: Cach
...
-              return toBeRemoved[idField] !== item[idField];
-            });
-          } else if (queryResult[queryField].items) {
-            const newItems = queryResult[queryField].items.filter((item: any) => {
+
+          const prev = (returnField)
+            ? queryResult[queryField][returnField]
+            : queryResult[queryField];
+
+          if (typeof prev.filter === "function") {
```
wtrocki, line 258: When I think about it - it will be better to check if prev is array

kingsleyzissou: Okay sure, easy fix

docs/ref-cache.md
kingsleyzissou: Thanks, I struggled with this a little

PR #537, opened 2020-06-25
Update example to use offix boost

wtrocki:
Still WIP. Using this to reproduce issues and hoping to fix one :D

### DIFF
--- a/examples/react/package.json
+++ b/examples/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "offix-react-example",
-  "version": "0.15.0",
+  "version": "0.15.4",
   "main": "index.js",
   "license": "Apache-2.0",
   "private": true,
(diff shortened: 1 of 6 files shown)


wtrocki:
This is WIP, but want to know about what you think of us using offix boost in example

wtrocki:
Screenshot 2020-06-25 at 11 55 43

Some errors happening with sample app - No matter if this branch or master

kingsleyzissou:
Yeah we can use offix-boost, no problem for me there.

The error is a bit strange. Not getting it on my end. But will double check

wtrocki:
Update. Offix releases have some code that is not on the master.

onStatusChangeListener is still there but it is no longer supported. 
React-offix-hooks package:
Screenshot 2020-06-25 at 12 29 30

wtrocki:
Used this PR to check file integration. We do not have file support in graphql-serve so we cannot apply this.

PR #544, opened 2020-06-29
Use offix-datastore in example apps

Eunovo:
### Description

This PR will contain work that modifies all the example apps to use offix-datastore. The react(web) example app has already been modified in this PR to use offix-datastore, discarding apollo and other offix libraries. Work on the example apps for mobile is on hold because offix-datastore doesn't support mobile devices yet.

##### Checklist

- [x] `npm run build` works


kingsleyzissou:
@Eunovo just going through this now. Are the changes reflected immediately for you? Or only on refresh?

Eunovo:
Sorry @kingsleyzissou I forgot to mention that the feature to listen for changes on datastore is currently broken. So, you can only see changes on refresh. I'll do something about this today

examples/react/src/hooks.js
```diff
@@ -0,0 +1,95 @@
...
+        })();
+    }, []);
+
+    return {
+        ...state, data: state.data ? state.data : []
+    }
+}
+
+export const useAddTodo = () => {
+    const [state, dispatch] = useReducer(reducer, initialState);
```
kingsleyzissou, line 50: Might be better to use context and do the `useReducer` only once, no?

This line could then become `const {state, dispatch} = useContext(CONTEXT_NAME);` and that way we are not re-creating the reducer multiple times with the `initialState`.

I would also almost be tempted to change things up here so that you store the model queries in the initialState. i.e. If you had `todos` and `comments`, your initial state could be 

```
const initialState = { todos: [], comments: [] }`
```

And then your reducer dispatch switch statements to handle the different CRUD types, eg:

```
async function reducer(state, action) {
    switch (action.type) {
        case Actions.FIND_TODOS:
			const todos = await TodoModel.query();
            return { ...state, loading: true, todos };
       	...
        default:
            throw new Error("Invalid action");
    }
}
```

This is just a suggestion. It may invalidate the need for some of the hooks though.

kingsleyzissou: Not too sure if you can do async reducer functions though.

wtrocki:
Awesome. Is this working - happy to check this out for you

Eunovo:
@wtrocki the refresh problem has been resolved. I can't run graphql-serve in `examples/react` though because of multiple graphql modules in same project. I'm currently trying to resolve this, so I can see the "ADD" operation replication in action

examples/react/src/hooks.js
Eunovo: @kingsleyzissou are you talking about storing todos in a global context and then dispatching actions to it? If you are, how should users make unique queries with the useFindTodos hook? Note that, I can easily allow users to pass parameters to be used for querying into the hook state by changing the code a little.

Eunovo:
@wtrocki You can test it. I have seen the "ADD" operation replication work in example app.

Eunovo:
"UPDATE" and "DELETE"(technically a special "UPDATE") are not working yet

wtrocki:
as for graphql-serve use npm with 0.14.0-beta2

examples/react/src/hooks.js
kingsleyzissou: My concern is if we have an app that has 4 or 5 or n amounts of models, we will end up having a `find`, `create`, `update` and `delete` hook for each case. This would become unmanageable pretty quickly. Whereas a more generic hook, might work better. This is just a suggestion though as this is part of a sample app and not hooks that we are providing the developer.

Eunovo: We can make the hooks more generic by passing the Model as an input parameter
`useAdd(TodoModel);`
This we can create only one hook for each operation that are reusable for all models

wtrocki:
@Eunovo Can we get that into next branch. I would love to see us working on next branch to identify blockers etc.

Eunovo:
@wtrocki  We can merge once you approve

examples/react/src/hooks.js
wtrocki: My suggestion is to create issue with this details. `useAdd` or `useDataStore(Action.ADD)` or `useDataStore().add()` as options to talk about

wtrocki:
I still saw this as draft. Approved

examples/react/src/hooks.js
wtrocki: @Eunovo Can you create issue for this?

Issue #49, opened 2019-06-12
Build network state aware data replication using plain graphql client


wtrocki:
I think the best way to start with this is simply to enable the application to Query specific data on the server and Subscribe for results when:

- Application is starting 
- The application became online and it is on the foreground. 

We currently have `OfflineMutationsHandler` that gives the capability to resend offline mutations. 

## Proposal for client

The client can have new methods for registering queries/subscriptions:
```
client.registerOnlineQuery(new OnlineQuery({gql,variables}))
client.registerOnlineSubscription(new OnlineSubscription({gql,variables}))
```

`OnlineQuery/OnlineSubscription` apart from having all required fields to perform mutation will contain metadata used to decide when to call mutate.

For example:
```typescript
// Wait with request after becoming online
public initialDelay: number = 0; 
// Interval used for pooling
public interval?: 
// Even some extra metadata
public requiresWifi: boolean
```
Developers will be able to trigger Query Refresh manually (and force subscriptions to reconnect:
```
client.forceOnlineRefresh();
```

## Related work
0. Extend OfflineMutationsHandler to handle subscriptions
1. Create abstraction for OnlineQuery and OnlineSubscription
2. Create Registry of OnflineQueryes/Subscriptions.
Expose methods to register OnlineQueries and OnlineSubscriptions to registry
3. Connect NetworkState interface to interact with the registry and execute depending on medatadata in  OnlineQuery and OnlineSubscription
4. Write unit tests and integration tests for this feature.

**Open for comments, opinions and contributions**
We can create individual issues once the community will agree on the flow.

Discussion in #6 for more information

wtrocki:
@Eunovo Let;s chat about this this week. Happy to explain this concept.

Eunovo:
> We currently have OfflineMutationsHandler that gives the capability to resend offline mutations.

@wtrocki you want to build client that uses this, right?

Eunovo:
Current offix client can handle offline mutations; You want to be able to handle subscriptions?

wtrocki:
Forget to talk about it. Lets ignore that for the moment but bring it to the project

Eunovo:
All right

Eunovo:
@wtrocki data replication between Apollo's single key store(local) and server or we get started on DataStore and use our own store.
I'm kinda confused, which are we doing? I'm thinking DataStore but this issue is old and talks about using apollo

Eunovo:
[...]

PR #550, opened 2020-07-01
Datastore replication engine

Eunovo:
### Description

Prototype Replication Engine which uses Offix-Scheduler to queue mutations. Currently, only "ADD" events can be replicated. "UPDATE" and "DELETE" events will probably require the generation of filter inputs to be sent with inputs to server. Also, proper error handling is still pending.

See #49 for discussions on this work.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


Eunovo:
@kingsleyzissou  We have bundle size causing tests to fail.

packages/offix-datastore/src/replication/ReplicationEngine.ts
```diff
@@ -0,0 +1,42 @@
...
+    private storage: Storage;
+
+    constructor(
+        api: IReplicator,
+        storage: Storage
+    ) {
+        this.api = api;
+        this.storage = storage;
+        this.scheduler = new Promise((resolve, reject) => {
+            const scheduler = new OffixScheduler<StoreChangeEvent>({
```
wtrocki, line 21: This looks really ugly. We need scheduler to change. There is no need for init for scheduler in new API etc.

wtrocki: Again this is more on core team shoulders to improve scheduler. Gianluca works on this and we can see API changed.

wtrocki:
Let's disable size checks for moment on this repo

kingsleyzissou:
@Eunovo @wtrocki we can disable the size-limit stuff, but it's actually quite easy to update now that I've added the config. Eunovo, you can go into the `.size-limit.js` file and just update the config file and set a buffer.

packages/offix-datastore/src/replication/ReplicationEngine.ts
kingsleyzissou: Yeah agreed, this is something that we've needed to improve for a while now.

packages/offix-datastore/src/replication/GraphQLCrudQueryBuilder.ts
```diff
@@ -0,0 +1,49 @@
...
+            const fieldsBuilder: string[] = Object.keys(fields).map((key) => {
+                const graphQLKey = fields[key].key;
+                return graphQLKey;
+            });
+            const graphQLFields = fieldsBuilder.join("\n");
+            const modelName = model.getName();
+
+            const mutations = {
+                create: gql`
+                mutation create${modelName}($input: Create${modelName}Input!) {
```
kingsleyzissou, line 25: @Eunovo @wtrocki we should probably have some documentation to explain that this is aligned to the graphqlcrud spec.

wtrocki: I would use this as a prototype for the moment. We winning and getting lots of progress. Once we get prototype working end to end with all we want to have, we will need to smooth it up and document it

Eunovo: Sure, for now, I only add a jsDoc comment on the class to indicate it's purpose

Eunovo:
There's a lot of work to be done here, but we can merge this into next. I'm creating new issues for other features

wtrocki:
Yes. I agree that this is still not prod ready but we should kinda work on individual TODO's together. Amazing work

PR #554, opened 2020-07-02
Fix problem with multiple subscribers

Eunovo:
### Description

Previous subscribers to Storage stream are overwritten when a new one is added. I have added a fix that collects all observers in an array which solves the problem. I update the "Observe local store for changes" test in `DataStore.test.ts` to include two subscriptions to verify that the fix works. 

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


Eunovo:
@wtrocki

wtrocki:
good to merge. Small change. Will review on feature branch

PR #561, opened 2020-07-07
fix: allow to use offline change without optimistic response

wtrocki:
Even driven engine messes things up - it tries to restore changes that replication already removed. 

Screenshot 2020-07-07 at 16 20 11

Screenshot 2020-07-07 at 16 20 09

This fix patches the problem to avoid breaking client. However entire event driven locking need to be done - in separate PR


kingsleyzissou:
I actually got this error at one point, but then couldn't replicate it again.

wtrocki:
It is related to event driven approach - easy to replicate if you have dangling operation that fails on server and gets removed from queue but then it is tried to be applied to db. 

Those are the current issues in the offix, but it looks like that datastore and 1.5 will have the same problems if we going to rely on events internally without awareness. I'm **ok** with using events for certain things as long we do not end up with stuff like this

wtrocki:
CC @Eunovo

PR #562, opened 2020-07-07
fix: add last updated state 

wtrocki:
### Description

Adding last updated state to be part of the supported states. 
No docs yet - need to think if we want to swap from versioned state etc. as default.


kingsleyzissou:
Just the linting :)

PR #563, opened 2020-07-07
Push Update and Delete operations to server

Eunovo:
### Description

Push Update and Delete operations to the server. Since update and delete are executed in batches, each affected row is pushed as a single operation to the server.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/offix-datastore/src/DataStore.ts
+++ b/packages/offix-datastore/src/DataStore.ts
@@ -35,7 +35,7 @@ export class DataStore {
         this.url = config.url;
         this.models = [];
     }
-
+    // TODO create model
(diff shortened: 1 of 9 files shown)


Eunovo:
@wtrocki  @kingsleyzissou  Update op replication fails when tested in example app. It seems there seems to be an error with the id

kingsleyzissou:
Okay will have a look at this. Which id is being sent? Is it the uuid generates by the data store or the server generated Id?

Eunovo:
The uuid generated by the datastore

wtrocki:
@Eunovo Can we fix build and merge this. 
I would love to rebase the next branch and remove some of the non datastore changes.

PR #586, opened 2020-07-09
Typesafe and complete operators closes #580

### LINKED ISSUE #580: Filter Operators for DataStore
## Feature Request

The Operators currently available in DataStore doesn't contain all or enough of GraphQLCrud operators. 

These operators are also not typesafe.
For example; you can do this `p.anyField("gt", 5)` or `p.anyField("gt", "random string")`  in current DataStore without any compile time error.

Two things are wrong here;
- The "gt" operator should only be available on number fields
- The "gt" operator should only accept numbers as inputs

cc @wtrocki  @kingsleyzissou 

Edit:
See:
https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/runtime/QueryFilter.ts

https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/client/src/dataFacade.tsx

Eunovo:
### Description

Implemented all the operators used in [GraphQLCrud filters](https://graphqlcrud.org/docs/next/find/#filtering). I modified Predicates such that the type of a field determines the allowed operators for that field. For example, numbers cannot use the "startsWith" operator. Also, only values of the same type or type[] can be compared. For example, you can perform the greater than operation on two numbers, not a number and a string.
 
##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works

### DIFF
--- a/packages/offix-datastore/src/predicates/Operators.ts
+++ b/packages/offix-datastore/src/predicates/Operators.ts
@@ -1,10 +1,40 @@
-const equalsOperator = (modelField: any, value: any) => modelField === value;
-const gtOperator = (modelField: any, value: any) => modelField > value;
+type CommonOperators = 'ne' | 'eq' | 'le' | 'lt' | 'ge' | 'gt';
 
-// https://github.com/aerogear/OpenVolunteerPlatform/blob/master/platform/client/src/dataFacade.tsx
-// TODO ADD type safety
-// TODO align operators to graphqlcrud
-export const AllOperators: any = {
-    "eq": equalsOperator,
-    "gt": gtOperator
+/**
+ * All Operators
+ * 
+ * 'ne' - Is value not equal to input
+ * 'eq' - Is value equal to input
+ * 'le' - Is value less than or equal to input
+ * 'lt' - Is value strictly less than input
+ * 'ge' - Is value greater than or equal to input
+ * 'gt' - Is value strictly greater than input
+ * 'in' - Does input array or string contain value
+ * 'contains' - Is the input contained in value(array or string)
+ * 'startsWith' - Does value start with input string
+ * 'endsWith' - Does value end with input string
+ */
+export type AllOperators = CommonOperators | 'in' | 'contains' | 'startsWith' | 'endsWith';
+
+/**
+ * Maps data type to allowed operators
+ */
+export type TypeOperatorMap<T> = 
+    T extends string ? AllOperators :
+    T extends number ? CommonOperators :
+    T extends boolean ? 'ne' | 'eq' :
+    T extends Date ? CommonOperators :
+    AllOperators;
+
+export const OperatorFunctionMap = {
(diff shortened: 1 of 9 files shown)


packages/offix-datastore/src/predicates/Operators.ts
```diff
@@ -1,10 +1,45 @@
-const equalsOperator = (modelField: any, value: any) => modelField === value;
-const gtOperator = (modelField: any, value: any) => modelField > value;
+type ArrayOperators = "ne" | "eq" | "in" | "contains";
```
wtrocki, line 1: Once again. Would be best to use enums here?

packages/offix-datastore/src/predicates/Operators.ts
```diff
@@ -1,10 +1,45 @@
...
+    T extends boolean[] ? ArrayOperators:
+    T extends boolean ? "ne" | "eq" | "in" :
+    T extends Date[] ? ArrayOperators:
+    T extends Date ? MathematicalOperators :
+    AllOperators;
+
+export const OperatorFunctionMap = {
+    ne: (m: any, v: any) => m !== v,
+    eq: (m: any, v: any) => m === v,
+    ge: (m: number | string | Date, v: number | string | Date) => m >= v,
```
wtrocki, line 37: Really nice workaround for now. We can make this generic by accepting type itself. Something like here: https://github.com/aerogear/graphback/blob/master/packages/graphback-core/src/runtime/QueryFilter.ts#L82

wtrocki:
@Eunovo how did you tested this knowing that current app is JS. Any trick I can do to check how typing are resolved in vscode?

Eunovo:
Use the test files. DataStore.test.ts as a lot of predicates in it. You can check what types are resolved by playing with the code there

wtrocki:
Actually enums here make less sense.
Screenshot 2020-07-09 at 15 59 40

wtrocki:
Challenge here is that we cannot predict invalid predicate - like using int with string operation and throw error. I see this as good work but we definitely need follow up. Since we pass models we can actually add additional field into model that will check type and validate queries at runtime. 

WDYT @Eunovo

Eunovo:
@wtrocki  don't merge yet. I wanna try something

Eunovo:
> Challenge here is that we cannot predict invalid predicate - like using int with string operation and throw error. I see this as good work but we definitely need follow up. Since we pass models we can actually add additional field into model that will check type and validate queries at runtime.

@wtrocki We can pull this off but I'm thinking about the cost of doing this. Is runtime validation worth the effort? Do we need to verify predicates? What would happen if we didn't?

wtrocki:
I mean this is stretch goal. What we have now it ok and we can improve this by generating typescript models or something - similar to generated schema typings from grpahql CRUD. I'm happy to merge this

Eunovo:
@wtrocki I have fixed conflicts with master. You can merge

PR #587, opened 2020-07-09
Update examples ts

kingsleyzissou:
### Description

I've updated both sample apps to `create-react-app` and `TypeScript` so we can properly test the datastore types. The update of the previous sample app was so we could have a side-by-side comparison if we need.

I have found a couple of issues that I will log separately.

There seems to be a memory leak in the datastore app, I think it has something to do with the hooks, but I didn't have enough time to properly bug fix.

The second issue is the graphql mutation input types are incorrect in the replication engine, causing the replication to fail.

I also changed the styles to `antd` and brought in `uniforms` and made a couple of improvements to the demo.


wtrocki:
I have couple comments and improvements but that will be easier to work on top of master. 
Tried it locally and already basing on it so need that on master.

PR #602, opened 2020-07-13
Receive changes from server

Eunovo:
### Description

This PR prepares us for MVP1. Here's what I've done here:
- I've changed the Model API so we can accept more configurations
- Given the Model objects access to a global replicator
- The previous two items allowed me to start delta synchronization in the Model object
- I stubbed out Subscriptions in the Model class

What we need to get delta synchronization working:
- we need a metadata store to hold lastSync for models
- we need a Predicate to GraphQLCrud filter converter
- we need to handle errors (technically the whole package needs this as there is no error handling yet) (Conflicts, etc.)

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/offix-datastore/src/Model.ts
+++ b/packages/offix-datastore/src/Model.ts
@@ -1,5 +1,6 @@
 import { Storage, StoreChangeEvent, DatabaseEvents } from "./storage";
 import { createPredicate, Predicate } from "./predicates";
+import { IReplicator } from "./replication";
 
 export interface FieldOptions {
     /** GraphQL type */
@@ -23,7 +24,8 @@ export type Fields<T> = {
 export interface ModelConfig<T = unknown> {
     name: string,
     storeName?: string,
-    fields: Fields<T>
+    fields: Fields<T>,
+    predicate?: Predicate<T>,
 }
 
 /**
@@ -37,12 +39,17 @@ export class Model<T = unknown> {
 
     constructor(
         config: ModelConfig<T>,
-        getStorage: () => Storage
+        getStorage: () => Storage,
+        replicator?: IReplicator
     ) {
         this.name = config.name;
         this.storeName = config.storeName || `user_${config.name}`;
         this.fields = config.fields;
         this.getStorage = getStorage;
+
+        if(replicator) { this.getDelta(replicator, config.predicate); }
+        // TODO remove ReplicationEngine 
+        // and push changes to replicator from change methods(CUD) here instead
     }
 
     public getFields() {
@@ -95,5 +102,27 @@ export class Model<T = unknown> {
                 if (event.eventType !== eventType) { return; }
                 listener(event);
             });
+    }
+
+    private async getDelta(replicator: IReplicator, predicate?: Predicate<T>) {
+        // TODO limit the size of data returned
+        const data = await replicator.pullDelta(this.name, "", predicate);
+
+        data
+        .filter((d: any) => (d._deleted))
+        .forEach((d: any) => this.remove((p: any) => p.id("eq", d.id)))
+
+        data
+        .filter((d: any) => (!d._deleted))
+        .forEach(async (d: any) => {
+            // TODO Predicate Matcher should be defined in config by user
+            const results = await this.update(d, (p: any) => p.id("eq", d.id));
+            if (results.length === 0) {
+                // no update was made, save the data instead
+                this.save(d);
+                return;
+            }
+        });
+        // TODO consider removing older data if local db surpasses size limit
     }
 }
--- a/packages/offix-datastore/src/Model.ts
+++ b/packages/offix-datastore/src/Model.ts
@@ -146,17 +146,19 @@ export class Model<T = unknown> {
 
         data
             .filter((d: any) => (d._deleted))
-            .forEach((d: any) => this.remove(matcher(d)));
+            .forEach((d: any) => { this.remove(matcher(d)); });
 
         data
             .filter((d: any) => (!d._deleted))
-            .forEach(async (d: any) => {
-                const results = await this.update(d, matcher(d));
-                if (results.length === 0) {
-                    // no update was made, save the data instead
-                    this.save(d);
-                    return;
-                }
+            .forEach((d: any) => {
+                (async () => {
+                    const results = await this.update(d, matcher(d));
+                    if (results.length === 0) {
+                        // no update was made, save the data instead
+                        this.save(d);
+                        return;
+                    }
+                })();
             });
         // TODO replicator.pullDelta should return lastSync and write to metadata store for model
         // TODO consider removing older data if local db surpasses size limit
--- a/packages/offix-datastore/src/Model.ts
+++ b/packages/offix-datastore/src/Model.ts
@@ -22,9 +22,24 @@ export type Fields<T> = {
  * Model Config options
  */
 export interface ModelConfig<T = unknown> {
+    /**
+     * Model name
+     */
     name: string,
+
+    /**
+     * Model store name, defualts to `user_${name}`
+     */
     storeName?: string,
+
+    /**
+     * Model fields
+     */
     fields: Fields<T>,
+
+    /**
+     * Delta Query and Subscription filter
+     */
     predicate?: Predicate<T>,
 
     /**
@@ -56,9 +71,12 @@ export class Model<T = unknown> {
         this.fields = config.fields;
         this.getStorage = getStorage;
 
+        // Model has access to replicator so there is possibility of model specific replicator config
+        // We can control Model specific replication here
         // TODO set default matcher
-        if(replicator && config.matcher) {
-            this.doDeltaSync(replicator, config.matcher, config.predicate,);
+        if (replicator && config.matcher) {
+            this.subscribeForServerEvents(replicator);
+            this.doDeltaSync(replicator, config.matcher, config.predicate);
         }
         // TODO remove ReplicationEngine 
         // and push changes to replicator from change methods(CUD) here instead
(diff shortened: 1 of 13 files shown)


Eunovo:
@wtrocki  @kingsleyzissou  I think we can build on this structure.

Eunovo:
Actually, there is one more thing missing. I haven't passed the replicator to the Model in DataStore.create. I'll do that tomorrow.

packages/offix-datastore/src/Model.ts
```diff
@@ -17,6 +18,40 @@ export type Fields<T> = {
...
+    predicate?: Predicate<T>;
+
+    /**
+     * Associate server version of entities with local version.
+     * It returns a predicate used to determine which entities
+     * get changed given the data from the server
+     *
+     * @param data is the version from the server
+     */
+    matcher?: (data: T) => Predicate<T>;
```
wtrocki, line 52: I'm not really getting what this does? Should we simply have two fields instead:

- subscriptionPredicate
- deltaQueryPredicate

packages/offix-datastore/src/Model.ts
```diff
@@ -17,6 +18,40 @@ export type Fields<T> = {
...
+ * Model Config options
+ */
+export interface ModelConfig<T = unknown> {
+    /**
+     * Model name
+     */
+    name: string;
+
+    /**
+     * Model store name, defualts to `user_${name}`
```
wtrocki, line 31: `data_${name}` seems to make more sense (and be more innovative)

packages/offix-datastore/src/Model.ts
```diff
@@ -17,6 +18,40 @@ export type Fields<T> = {
...
+
+    /**
+     * Model fields
+     */
+    fields: Fields<T>;
+
+    /**
+     * Delta Query and Subscription filter
+     */
+    predicate?: Predicate<T>;
```
wtrocki, line 43: Delta query predicate can be different than subscription (due to subscription limitations. We do not have this well documented so just mentioning it now for the moment. Subscriptions are still not well covered in backend (yet)

packages/offix-datastore/src/Model.ts
```diff
@@ -27,15 +62,24 @@ export class Model<T = unknown> {
...
-        getStorage: () => Storage
+        config: ModelConfig<T>,
+        getStorage: () => Storage,
+        replicator?: IReplicator
     ) {
-        this.name = name;
-        this.storeName = storeName;
-        this.fields = fields;
+        this.name = config.name;
+        this.storeName = config.storeName || `user_${config.name}`;
```
wtrocki, line 70:
```suggestion
        this.storeName = config.storeName || `data_${config.name}`;
```

packages/offix-datastore/src/Model.ts
```diff
@@ -89,4 +133,34 @@ export class Model<T = unknown> {
                 listener(event);
             });
     }
+
+    private async subscribeForServerEvents(replicator: IReplicator) {
```
wtrocki, line 137: I like that we have it globally available, but WDYT if we will have it public predicate instead. 

This will mean that user can:

- Add predicate generically on init 
- Subscribe to server events in specific view only (without persisting this predicates or reatempting this on startup. 

Not sure if that makes sense for now, but would love if you can tell me what you think about this.

packages/offix-datastore/src/Model.ts
```diff
@@ -89,4 +133,34 @@ export class Model<T = unknown> {
...
+    }
+
+    private async doDeltaSync(replicator: IReplicator, matcher: (d: T) => Predicate<T>, predicate?: Predicate<T>) {
+        // TODO limit the size of data returned
+        // TODO get lastSync for model for metadata store and pass to pullDelta
+        const data = await replicator.pullDelta(this.getStoreName(), "", predicate);
+
+        data
+            .filter((d: any) => (d._deleted))
+            .forEach((d: any) => { this.remove(matcher(d)); });
```
wtrocki, line 149: I would like to know why we doing this? What is the use case.

packages/offix-datastore/src/Model.ts
```diff
@@ -89,4 +133,34 @@ export class Model<T = unknown> {
...
+
+    private async subscribeForServerEvents(replicator: IReplicator) {
+        // TODO replicator.subscribe
+        // handle subscription events here
+    }
+
+    private async doDeltaSync(replicator: IReplicator, matcher: (d: T) => Predicate<T>, predicate?: Predicate<T>) {
+        // TODO limit the size of data returned
+        // TODO get lastSync for model for metadata store and pass to pullDelta
+        const data = await replicator.pullDelta(this.getStoreName(), "", predicate);
```
wtrocki, line 145: [Question] How we avoid user app being bombarded by many events when multiple items were returned (this is different from subscriptions that are only per single item)

packages/offix-datastore/src/Model.ts
```diff
@@ -89,4 +133,34 @@ export class Model<T = unknown> {
...
+
+        data
+            .filter((d: any) => (d._deleted))
```
wtrocki, line 148: [Discussion] I think we should still save deleted item for conflict resolution - purposes. Without that we will need to force deltas for every modification that happens on the server as we can have local conflicts.

packages/offix-datastore/src/replication/GraphQLCrudQueryBuilder.ts
```diff
@@ -43,7 +43,9 @@ export class GraphQLCrudQueryBuilder implements GraphQLQueryBuilder {
                 }`
       };
 
-      const queries = {
+      const queries: ReplicatorQueries = {
+        // TODO create a GraphQLCrudQuery class implementing ReplicatorQuery
```
wtrocki, line 47: +1

FYI - GraphQL crud will deal with the mapping for different providers in near future but having offix working with any user-provided server would be amazing +100 feature

packages/offix-datastore/src/replication/GraphQLCrudQueryBuilder.ts
```diff
@@ -55,15 +57,19 @@ export class GraphQLCrudQueryBuilder implements GraphQLQueryBuilder {
...
+            query sync${modelName}($lastChanged: String!, $filter: ${modelName}Filter) {
+              find${modelName}s(lastChanged: $lastChanged, filter: $filter) {
+                  items {
+                    ${graphQLFields}
                   }
-                }`,
+                  lastChanged
+              }
+            }`,
+            getData: (response) => (response.data.items),
```
wtrocki, line 70: -1 to have this two methods. I guess we will get that from the mentioned class

packages/offix-datastore/src/replication/GraphQLReplicator.ts
```diff
@@ -33,35 +34,55 @@ export interface ReplicatorMutations {
...
-  updated: string | DocumentNode;
-
-  /**
-   * GraphQL delete subscription document.
-   */
-  deleted: string | DocumentNode;
+    /**
+     * GraphQL create subscription document.
+     */
+    new: string | DocumentNode;
```
wtrocki, line 40: actually `new` is a keyword so we might avoid this name :) 
My bad!

packages/offix-datastore/src/replication/GraphQLReplicator.ts
```diff
@@ -33,35 +34,55 @@ export interface ReplicatorMutations {
...
+    updated: string | DocumentNode;
+
+    /**
+     * GraphQL delete subscription document.
+     */
+    deleted: string | DocumentNode;
+}
+
+// TODO add predicateToFilter
+interface ReplicatorQuery {
```
wtrocki, line 54: This seems redundant - What would be the reason for this interface (and one bellow)

packages/offix-datastore/src/replication/GraphQLReplicator.ts
```diff
@@ -126,4 +154,20 @@ export class GraphQLReplicator implements IReplicator {
...
     }
+
+    public async pullDelta<T>(storeName: string, lastSync: string, predicate?: Predicate<T>) {
+        const syncQuery = this.queries.get(storeName)?.queries.sync;
+        if (!syncQuery) {
+            throw new Error(`GraphQL Sync Queries not found for ${storeName}`);
+        }
+        /**
+         * TODO convert predicate to filter here.
+         * The syncQuery object should do this conversion as it is from the GraphQLCrud Layer
```
wtrocki, line 165: +1

wtrocki:
Amazing work. 
I will add couple TODO's here on top of what we have

1) Model user events (currently in-store) to be able to receive updates for specific predicate.

Subscription and delta can return some specific general dataset data (multiple items).
This means that when we are on the list we do not want to be bombarded with multiple events (or require users to aggregate events in 2-3 second batches - rxjs looking at you here)

It is probably the best to enable multiple models to be returned at once and aggregate those thru replication engine .

Another thing is to be able to watch for specific predicate - like getting user details view we can watch for subscription changes on this particular screen while we will be getting the updates for all the objects in the background. 

2) Subscriptions and Deltas might need to be split into two separate engines. Not everyone would have ability to do subscriptions. Deltas are really powerful and for some of the use cases they will be efficient. Specifying explicitly this will allow us to reduce size of the package (entire subscriptions part can go away. Archives better code separation and have the simplicity of integrating with various dev use cases.

This PR did not affect much for me and I should be able to based on this to add mutation replication as well

packages/offix-datastore/src/Model.ts
Eunovo: Really anything you like

packages/offix-datastore/src/Model.ts
Eunovo: No problem. Add separate predicates for subscription and delta query

packages/offix-datastore/src/Model.ts
Eunovo: Remember that we intend to move events.
Here, we would not use normal model crud methods(which emits events per action) but instead call storage directly

packages/offix-datastore/src/replication/GraphQLCrudQueryBuilder.ts
Eunovo: Yes. It's a TODO

packages/offix-datastore/src/Model.ts
Eunovo: I don't understand what you mean here. Are you asking about the matcher?

packages/offix-datastore/src/replication/GraphQLReplicator.ts
Eunovo: Won't be a problem.

packages/offix-datastore/src/replication/GraphQLReplicator.ts
Eunovo: Response formats for queries vary. CRUD Spec uses `result.items`. I'm using this interface to avoid forcing Crud Spec in the `GraphqlReplicator` layer. The one below is only for sync queries. It was just easier to define its own interface as it has a lastSync response we want to access, not just items

packages/offix-datastore/src/Model.ts
wtrocki: Why we remove `_deleted` elements?

PR #609, opened 2020-07-14
fix: make sample app work

wtrocki:


examples/react-datastore/src/components/forms/AddTodo.tsx
```diff
@@ -17,6 +17,7 @@ export const AddTodo = ({ cancel }: AddTodoProps) => {
 
   const handleSubmit = ({ title, description }: ITodo) => {
     addTodo({
+      id: "whatever",
```
wtrocki, line 20: @Eunovo this is probably not wanted side effects of typings.

PR #610, opened 2020-07-14
fix: patch current types and add temporary id generation

wtrocki:
Patching things up. Current api is to strongly typed and requires us to have 2 variations of typing for each operation.

### DIFF
--- a/examples/react-datastore/package.json
+++ b/examples/react-datastore/package.json
@@ -24,13 +24,15 @@
     "typescript": "~3.7.2",
     "uniforms": "3.0.0-alpha.5",
     "uniforms-antd": "3.0.0-alpha.5",
-    "uniforms-bridge-graphql": "3.0.0-alpha.5"
+    "uniforms-bridge-graphql": "3.0.0-alpha.5",
+    "uuidv4": "^6.1.1"
   },
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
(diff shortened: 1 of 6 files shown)


examples/react-datastore/package.json
```diff
@@ -24,13 +24,15 @@
...
+    "uniforms-bridge-graphql": "3.0.0-alpha.5",
+    "uuidv4": "^6.1.1"
   },
   "scripts": {
     "start": "SKIP_PREFLIGHT_CHECK=true react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "eject": "react-scripts eject"
+    "eject": "react-scripts eject",
+    "linkdatastore": "cd ../../packages/offix-datastore && npm link . && cd - && npm link offix-datastore"
```
wtrocki, line 35: Added helper to link datastore

packages/offix-datastore/package.json
```diff
@@ -16,6 +16,7 @@
     "build": "tsc",
     "lint-fix": "eslint \"./src/**/*.ts\" \"./tests/**/*.ts\" --fix",
     "test": "jest",
+    "watch": "tsc --watch",
```
wtrocki, line 19: you can now watch changes so app will reload when you make any change in datastore

wtrocki:
Merging to fix issues on master - I need it for uncommented work in #606

PR #611, opened 2020-07-14
Datastore subscriptions

kingsleyzissou:
This is all very rough for now. There are a couple of issues I need to iron out.

But some of the basic logic has been added. Some of this has only sort of been stubbed out.


packages/offix-datastore/src/Model.ts
```diff
@@ -134,9 +134,15 @@ export class Model<T = unknown> {
             });
     }
 
-    private async subscribeForServerEvents(replicator: IReplicator) {
+    // TODO this method should be public
+    // so we can then specify the type of subscription event
+    // the replicator in this case should not be passed as a variable
+    private subscribeForServerEvents(replicator: IReplicator) {
```
kingsleyzissou, line 140: @wtrocki @Eunovo I want to make this method public and maybe change the name. The idea here is that you could subscribe at the app level in a specific view. The issue though is passing the `replicator` to the model.

kingsleyzissou: This issue is that the replication engine is initiated after the model has been created, so the as far as I can tell, the replicator is never going to be passed to the constructor method of the model

wtrocki: See what I have pointed in previous PR from @Eunovo 
This needs to be public and also needs to accept filter rather than replicator.

wtrocki:
awesome work

packages/offix-datastore/src/Model.ts
kingsleyzissou: @wtrocki yeah I left the replicator in for now, but I definitely plan on removing that and passing in the filter and the subscription type. Just not sure yet how we plan on making the replicator available to the model.

Eunovo: @kingsleyzissou If you do need to pass replicator to model, there are two things you could do;
- use a callback as I did with Storage
- refactor query builder to build queries, one model, at a time and add them to the replicator using a public function like `replicator.addQuery(storeName, query)`(which internally builds the query map) so that you could initialise the replicator before creating the models

kingsleyzissou: @Eunovo interesting, i her considered a callback function actually. I'll give it a try. 

Otherwise I was thinking of maybe having a getter and setter for the replicator, since the doDeltaSync method relies on it too.

Eunovo:
Yes. The getter and setter would work too

packages/offix-datastore/src/Model.ts
wtrocki: I think this makes no sense itself. Subscriptions are runtime setup only apparently. While delta makes sense to be forcefully called - we cannot really get live results from delta or subscriptions (all sadly needs to go thru the store - and notified to user back). 

I can do this quick fix, but this do not change that replication should be separate for subscriptions and deltas and play nice with store.

wtrocki:
I have been struggling with an extensive amount of abstractions and interfaces that are introduced too early.I see that we will need to start writing a replication engine from scratch it is very convoluted and obfuscates what we want to do. I wanted to do simple refactoring and ended up with 3 hours or moving types around - with poor results. There are multiple critical bugs with eventing, circular loops etc. I would love to focus on those first. Have robust storage - this actually makes sense as interface.

I have been trying to put simple MutationEventStore and DeltaMetadata but it will be quite hard to do it in the replication engine as it is. I would see the replication engine for the moment as two classes max with the embedded client etc. Due to extensive amount of interfaces, we cannot simply do error handling and other elements easily - we trading correctness over the abstractions.

PR #618, opened 2020-07-15
Datastore client

kingsleyzissou:
### Description
Remove client abstraction as we do not plan to provide support for other graphql clients at this present time


packages/offix-datastore/src/replication/graphqlcrud/GraphQLCRUDReplicator.ts
```diff
@@ -39,13 +42,13 @@ export class GraphQLCRUDReplicator implements IReplicator {
 
     switch (eventType) {
       case CRUDEvents.ADD:
-        return this.client.mutate<T>(mutations.create, { input });
+        return this.mutate<T>(mutations.create, { input });
```
wtrocki, line 45: This change will conflict with what I have build in other PR.

wtrocki:
Good work!

PR #617, opened 2020-07-15
Replication

wtrocki:
WIP replication engine.

TODO here is to connect it with the network status interface, but I did not want to do it now.

PR #621, opened 2020-07-16
Clientstore network

kingsleyzissou:
### Description
I have created a network status interface that uses the websocket connection as an indication of whether the client is online or offline. There are still a few todos and things that need to be added once the replication pr is merged (https://github.com/aerogear/offix/pull/617)

### DIFF
--- a/examples/react-datastore/src/App.tsx
+++ b/examples/react-datastore/src/App.tsx
@@ -2,17 +2,31 @@ import React, { useState, useEffect } from 'react';
 import { Button, Badge } from 'antd';
 import 'antd/dist/antd.css';
 
-import { TodoModel } from './config/datastoreConfig';
+import { datastore, TodoModel } from './config/datastoreConfig';
 import { useFindTodos } from './helpers/hooks';
 import { TodoList, AddTodo, Loading, Error, Header } from './components';
 import { CRUDEvents } from 'offix-datastore';
+import { NetworkEvent } from 'offix-datastore/types/utils/NetworkStatus';
 
 function App() {
 
   const [mounted, setMounted] = useState<boolean>(false);
   // TODO implement a network listener
+  const [isOnline, setIsOnline] = useState<boolean>(datastore.networkStatus.isOnline);
   const [addView, setAddView] = useState<boolean>(false);
   const  { loading, error, data } = useFindTodos();
+
+  // TODO create a hook for this
+  useEffect(() => {
+    datastore.networkStatus.subscribe((x: NetworkEvent) => {
+      console.log('network', x);
+      setIsOnline(x.status); 
+    });
+    return () => {
+      setIsOnline(false);
+      // TODO unsubscribe
+    }
+  }, [setIsOnline]);
 
   useEffect(() => {
     if (mounted) {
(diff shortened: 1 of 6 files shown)


kingsleyzissou:
@Eunovo @wtrocki I've exposed the network status and added an example usage of this to the sample app.

To verify:
- start the server
- start the client
- kill server to see indicator change to offline
- restart server (and wait) to see indicator change back to online

Still todo:
Add unsubscribe method to NetworkStatus class

examples/react-datastore/src/App.tsx
```diff
@@ -2,18 +2,32 @@ import React, { useState, useEffect } from 'react';
...
 
+  // TODO create a hook for this
+  useEffect(() => {
+    datastore.networkStatus.subscribe((x: NetworkEvent) => {
+      console.log('network', x);
+      setIsOnline(x.status); 
+    });
+    return () => {
+      setIsOnline(false);
+      // TODO unsubscribe
```
wtrocki, line 27: As agreed let's create follow up issue.

packages/offix-datastore/src/utils/NetworkStatus.ts
```diff
@@ -0,0 +1,38 @@
+import Observable from "zen-observable";
+import { SubscriptionClient } from "subscriptions-transport-ws";
+
+export interface NetworkEvent {
```
wtrocki, line 4: I will put some comments here

packages/offix-datastore/src/utils/NetworkStatus.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export class NetworkStatus {
+
+  public isOnline: boolean = false;
+  public subscriber: Observable<NetworkEvent>;
+
+  public constructor(subscriptionClient: SubscriptionClient) {
+    this.subscriber = new Observable((observer) => {
+      observer.next({ event: "initialising", status: false });
+      subscriptionClient.onConnected(() => observer.next({ event: "connected", status: true }));
```
wtrocki, line 23: Connected, dissconnected could be enums

wtrocki:
Having some unit tests here would be awesome although not essential to merge this change.
I have played with this an a got some issues with replication engine - but actually status events work ok - getting online offline statuses and they are independent from the implementation

packages/offix-datastore/src/utils/NetworkStatus.ts
```diff
@@ -0,0 +1,38 @@
...
+   * Network event type
+   */
+  event: string;
+  /**
+   * Connection status
+   */
+  status: boolean;
+};
+
+export class NetworkStatus {
```
wtrocki, line 15: Would we also drop some interface in future - maybe separate issue to allow people override that if needed? Def not target of the issue. Would that be possible to be introduced at some point?

kingsleyzissou: Yeah I will create an issue for this

kingsleyzissou: Actually, I just created and exposed an interface, this is a quick win. That and the enum.

The unsubscribing from the listener is a larger issue to tackle and we can do the tests along with this.

PR #628, opened 2020-07-17
Transaction support closes #625

### LINKED ISSUE #625: Transaction Support
## Feature Request

Transaction support for the storage layer

```
transaction = storage.createTransaction()
transaction.commit - fire all events, mutation replication starts
transaction.rollback - no events are fired
transaction.save(), .update, .delete
```
All events are queued until the transaction is committed

@wtrocki @kingsleyzissou

Eunovo:
### Description

Implement transaction api at Storage layer as described in #625.
To verify, see tests in `./tests/Storage.test.ts`

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


packages/offix-datastore/src/storage/LocalStorage.ts
```diff
@@ -82,4 +112,12 @@ export class LocalStorage {
   public async readMetadata(storeName: any): Promise<any> {
     return this.adapter.query(storeName);
   }
+
+  private dispatchEvent(event: StoreChangeEvent) {
+    if (!this.adapter.isTransactionOpen()) {
```
wtrocki, line 117: So technically it will allow multiple operations going thru transition as well. All controllable by Open and commit?

packages/offix-datastore/src/storage/LocalStorage.ts
```diff
@@ -17,15 +17,45 @@ export function generateId() {
...
+    const adapterTransaction = await this.adapter.createTransaction();
+    return new LocalStorage(adapterTransaction, this.storeChangeEventStream);
+  }
+
+  /**
+   * Commits a transaction if one is open.
+   * It fires all events that occured in this transaction
+   */
+  public async commit() {
+    if (!this.adapter.isTransactionOpen()) { return; }
```
wtrocki, line 41: This should throw really ugly error! 
Why. If this ever happens it usually will mean bug as you should not have code that commits transition that were not opened.

packages/offix-datastore/src/storage/LocalStorage.ts
```diff
@@ -17,15 +17,45 @@ export function generateId() {
...
+    await this.adapter.commit();
+    this.eventQueue.forEach((event) => this.storeChangeEventStream.push(event));
+  }
+
+  /**
+   * Rollback all changes that occured in this transaction, if one is open.
+   * All events that occured within this transaction are cleared.
+   */
+  public async rollback() {
+    if (!this.adapter.isTransactionOpen()) { return; }
```
wtrocki, line 51: Same as above. Throw error.

wtrocki:
Like how generic this is - it is not limited to metadata itself.

Eunovo:
@wtrocki  I have modified code to throw an error if commit or rollback is called outside a transaction

packages/offix-datastore/src/storage/LocalStorage.ts
Eunovo: yes

PR #632, opened 2020-07-20
Reapply changes

kingsleyzissou:
### Description
These are some of the changes that we had to revert from #617. We weren't able to get the tests to pass on this.

### DIFF
--- a/packages/offix-datastore/src/Model.ts
+++ b/packages/offix-datastore/src/Model.ts
@@ -172,15 +172,19 @@ export class Model<T = unknown> {
     if (data.data && data.data.length > 0) {
       data.data
         .filter((d: any) => (d._deleted))
-        .forEach((d: any) => { this.remove(matcher(d)); });
+        .forEach((d: any) => {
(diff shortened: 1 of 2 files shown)


packages/offix-datastore/src/Model.ts
```diff
@@ -172,15 +172,19 @@ export class Model<T = unknown> {
     if (data.data && data.data.length > 0) {
       data.data
```
kingsleyzissou, line 173: @wtrocki @Eunovo maybe we can extract this and the filter below to their own functions?
WDYT?

I think in this case it would make the code more readable.

wtrocki:
Now I see what you meant! Going to improve this PR

wtrocki:
Going to create separate PR

packages/offix-datastore/src/Model.ts
wtrocki: It is pointless to filter IMHO. we should have one module that:
- deduplicate
- deals with deletion (removes or simply have field as _deleted)
etc.

PR #639, opened 2020-07-21
Debugger

kingsleyzissou:
### Description
I've created a small helper file so we can have namespaced logging in the datastore. I haven't added it everywhere, but in a few places to show how it can be used.

The first step is to import the `createLogger` function.

Then to create a namespace `foo`, you would do the following:

```
const logger = createLogger('foo');
```

This creates a `logger` function which can be used to log events.

To enable it in the browser, we need to add:

```
localStorage.debu = 'datasync:*' // namespace for the store
```

### DIFF
--- a/packages/offix-datastore/src/storage/adapters/IndexedDBStorageAdapter.ts
+++ b/packages/offix-datastore/src/storage/adapters/IndexedDBStorageAdapter.ts
@@ -29,7 +29,7 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
     }
 
     public createStores(dbName: string, schemaVersion: number) {
-        logger("Creating stores");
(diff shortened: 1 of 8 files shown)


packages/offix-datastore/src/storage/adapters/IndexedDBStorageAdapter.ts
```diff
@@ -26,6 +29,7 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
     }
 
     public createStores(dbName: string, schemaVersion: number) {
+        logger("Creating stores");
```
wtrocki, line 32: It is best to include information about dbName and schemaVersion etc.

PR #654, opened 2020-07-24
Base architecture for DataSync

wtrocki:
Storage, and model will not change. I'm now investigating different approaches for the replication engine (avoiding event driven approach and splitting replication based on config etc. + exposing this to end user.


packages/offix-datastore/src/DataStore.ts
```diff
@@ -1,74 +1,77 @@
...
- * Configuration Options for DataStore
+ * Custom implementation
  */
-export interface DataStoreConfig {
-  /**
-   * The Database name
-   */
-  dbName: string;
-
+export interface CustomEngines {
```
wtrocki, line 20: Customization options for engines. Exposed that so we can support different replication and store. I'm simplifying things a lot. 

If you want to have different replication/client etc. you will need to swap entire replication (or extend ours) rather than building some swappable abstraction within replication itself.

packages/offix-datastore/src/DataStore.ts
```diff
@@ -1,74 +1,77 @@
...
-  private indexedDB: IndexedDBStorageAdapter;
   private storage: LocalStorage;
-  private clientConfig: any;
-  private metadataName = "metadata";
 
-  constructor(config: DataStoreConfig) {
-    this.dbName = config.dbName;
-    this.schemaVersion = config.schemaVersion || 1; // return 1 if schemaVersion is undefined or 0
-    this.clientConfig = config.clientConfig;
+  private config: DataStoreConfig;
```
wtrocki, line 44: Key for datastore was to have clear architecture.

Datastore should have only:
- Models
- Store
- Replication

We can connect them later.

packages/offix-datastore/src/DataStore.ts
```diff
@@ -1,74 +1,77 @@
...
-    this.indexedDB.createStores(this.dbName, this.schemaVersion);
-    engine.start();
-  }
+    this.storage.createStores(this.config);
 
-  private pushReplicator(replicator: IReplicator) {
-    this.models.forEach((model) => {
-      model.setReplicator(replicator);
-    });
+    if (this.config.replicationConfig) {
```
wtrocki, line 70: Replication works only when you pass config explicitly. 
Haven't decided it it starts automatically or requires separate method.

packages/offix-datastore/src/Model.ts
```diff
@@ -38,20 +38,6 @@ export interface ModelConfig<T = unknown> {
    * Model fields
    */
   fields: Fields<T>;
-
-  /**
-   * Delta Query and Subscription filter
-   */
-  predicate?: Predicate<T>;
```
wtrocki, line 45: Model should never actually have those as it mixing concerns - model represents local state and replication information should be passed separately.

packages/offix-datastore/src/Model.ts
```diff
@@ -76,7 +62,7 @@ export class Model<T = unknown> {
     this.fields = config.fields;
     this.storage = storage;
     this.metadataName = metadataName;
-    // TODO set custom primary keys
+    // TODO set primary keys here or thru api
     this.storage.adapter.addStore({ name: this.storeName });
```
wtrocki, line 66: I do not like that model executes and deals with its store. In my take model should be just stateless, immutable data container. Did not wanted to refactor this in this PR but that is must for the next phase IMHO

packages/offix-datastore/src/utils/logger.ts
```diff
@@ -3,3 +3,9 @@ import debug, { Debugger } from "debug";
 export function createLogger(name: string): Debugger {
   return debug(`datasync:${name}`);
 }
+
+export function enableLogger() {
+  if (window?.localStorage) {
+    window.localStorage.setItem('debug', 'datasync::*');
```
kingsleyzissou, line 9:
```suggestion
    window.localStorage.setItem('debug', 'datasync:*');
```

packages/offix-datastore/src/Model.ts
```diff
@@ -171,15 +151,19 @@ export class Model<T = unknown> {
     if (data.data && data.data.length > 0) {
       data.data
         .filter((d: any) => (d._deleted))
-        .forEach((d: any) => { this.remove(matcher(d)); });
+        .forEach((d: any) => {
```
wtrocki, line 154: This code will be removed - DeltaSync should not happen in the model. 
The model will just expose it's own replication proxy

packages/offix-datastore/src/network/NetworkStatus.ts
```diff
@@ -0,0 +1,24 @@
+/**
```
wtrocki, line 1: I have overly simplified NetworkStatus for the moment

packages/offix-datastore/src/network/SubscriptionNetworkStatus.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (this.isOnlineFlag === undefined) {
+      return super.isOnline();
+    }
+    return Promise.resolve(super.isOnline() && this.isOnlineFlag);
+  }
+
+  public subscribe(observer: ZenObservable.Observer<NetworkStatusEvent>) {
+      // Websockets seems to have disconnection lag
+      // TODO figure out best parameters for observability.
+      return super.subscribe(observer);
```
wtrocki, line 41: Challenge of using subscription based network status is that we need to have info from both resolvers. I will try to play with this on real device first and see what is the best option. 
Ideally we should rely on WebSocket but check window.network.isOnline before etc.

packages/offix-datastore/src/replication/GraphQLCRUDReplicator.ts
```diff
@@ -28,12 +33,23 @@ export function convertPredicateToFilter(predicate: PredicateFunction): any {
  * Performs replication using GraphQL
  */
 export class GraphQLCRUDReplicator implements IReplicator {
+
   private client: Client;
-  private queries: Map<string, GraphQLDocuments>;
+  // private networkStatus: NetworkStatus
```
wtrocki, line 38: Ignore that class completely and everything that is in the replicator. 
I have some 2 different sets of changes for this that I'm exploring.

packages/offix-datastore/src/replication/documents/buildGraphQLCRUDMutations.ts
```diff
@@ -0,0 +1,49 @@
+import gql from "graphql-tag";
+
+import { Model } from "../../Model";
+import { ReplicatorMutations } from "./Documents";
+
+/**
+ * Builds GraphQL mutations for models following
+ * the [GraphQLCRUD specification]{@link https://graphqlcrud.org/}
+ */
+export const buildGraphQLCRUDMutations = (models: Model[]): Map<string, ReplicatorMutations> => {
```
wtrocki, line 10: Since we can have a situation where we do not need subscriptions etc. so I have split creation into separate method for mutations, subscriptions and deltas. Those are build based on the config.

packages/offix-datastore/src/storage/LocalStorage.ts
```diff
@@ -116,6 +120,26 @@ export class LocalStorage {
     return this.adapter.query(storeName);
   }
 
+  /**
+   * Create store in underlying storage provider
+   *
+   * @param config
+   */
+  public addStore(config: IStoreConfig) {
```
wtrocki, line 128: LocalStorage is the only one element that people interact with at the moment. We can add store and createStores - but I see those methods here only temporarily. Store creation is specific to the IndexedDB etc.

I would drop SQL lite here and see how we can create stores on demand (this will make it hard for us to track the versions of the stores etc so current approach of creating all stores upfront and init is the best. 

Same as replication creation on stores is final and once app is initialized we cannot recreate it etc. (which makes this API undocumented and hard to use)

packages/offix-datastore/src/storage/adapters/IndexedDBStorageAdapter.ts
```diff
@@ -23,7 +23,7 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
         });
         this.transaction = transaction;
     }
-
+    // TODO Wrong architecture. Store can be created on demand
```
wtrocki, line 26: This is not exactly accurate but points out API problem. 
I hope to create issue with solution to not resolve too many problems in single PR.

packages/offix-datastore/src/storage/adapters/IndexedDBStorageAdapter.ts
```diff
@@ -56,6 +56,8 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
 
             this.stores.forEach(({ name, keyPath }) => {
                 if (existingStoreNames.contains(name)) { return; }
+                logger("Creating store", name, keyPath);
+                // TODO hardcoded id and unused keypath
```
wtrocki, line 60: Huge problem we have now is that we assume that id will be and `id` which is not going to work even by default as we move now to `_id` for Mongo in coming weeks. 

CC @ssd72

wtrocki:
Code will not compile, unit tests will fail and app will not work. 
Need entire day to enable app and going to ping you again. 
See my comments above.

packages/offix-datastore/src/utils/logger.ts
wtrocki: Yep.. I have pushed that by mistake. Have that changed.

packages/offix-datastore/src/DataStore.ts
```diff
@@ -1,76 +1,84 @@
...
+      if (this.engines?.replicator) {
+        this.engines?.replicator.start(this.models, this.storage);
+      } else {
+        const gqlReplicator = new GraphQLCRUDReplicator(this.config.replicationConfig);
```
kingsleyzissou, line 77: If we are planning to expose the replication engine, then maybe we can set the gqlReplicator as the default in the constructor if no replication engine is provided?

packages/offix-datastore/src/DataStore.ts
```diff
@@ -1,76 +1,84 @@
...
+        gqlReplicator.start(this.models, this.storage);
+      }
+    } else {
+      logger("Replication configuration was not provided. Replication will be disabled");
```
kingsleyzissou, line 81: Uhh okay. This answers previous comment

wtrocki:
While this PR doesn't has tests and subscriptions and other stuff I will be back working on it. 
Build and test will pass. We can exceptionally merge that stuff and continue while this will not be blocking anyone from doing improvements and causing tons of conflicts.

packages/offix-datastore/src/DataStore.ts
wtrocki: this is no longer valid. We need to wrap this later into different queries

packages/offix-datastore/src/network/SubscriptionNetworkStatus.ts
```diff
@@ -0,0 +1,40 @@
...
+import { NetworkStatusEvent } from "./NetworkStatus";
+import { WebNetworkStatus } from "./WebNetworkStatus";
+
+/**
+ * The default implementation of the NetworkStatus listener
+ * uses the websocket connection to the server to detect
+ * changes in the network status and use this as an indicator
+ * of the network status of the client.
+ */
+export class SubscriptionNetworkStatus extends WebNetworkStatus {
```
wtrocki, line 12: This is not used yet

packages/offix-datastore/src/replication/GraphQLReplicator.ts
```diff
@@ -0,0 +1,68 @@
...
+  delta: {
+    enabled: true,
+    pullInterval: 1000 * 60 * 1 // 1 minute
+  },
+  liveupdates: {
+    enabled: true
+  },
+  mutations: {
+    enabled: true
+  }
```
wtrocki, line 30: I forgot to add networkStatus

packages/offix-datastore/src/replication/ModelReplication.ts
```diff
@@ -0,0 +1,119 @@
...
+        query: queries.sync,
+        model: this.model
+      };
+      const replicator = new DeltaReplicator(deltaOptions);
+      replicator.start();
+    }
+
+    if (config.liveupdates?.enabled) {
+      // const subscriptionQueries = buildGraphQLCRUDSubscriptions(this.model);
+      // TODOs
```
wtrocki, line 53: Subscriptions replication is missing. Could not restore it

packages/offix-datastore/src/replication/api/ReplicationConfig.ts
```diff
@@ -0,0 +1,139 @@
...
+export interface LiveUpdatesConfig {
+  enabled: boolean;
+}
+
+/**
+ * Handle errors repeat request if needed
+ *
+ * @returns MutationRequest if request should be repeated
+ */
+export type UserErrorHandler = (networkError: any, graphqlError: any) => MutationRequest | undefined;
```
wtrocki, line 28: Should be renamed to MutationErrorHandler

packages/offix-datastore/src/replication/mutations/MutationRequest.ts
```diff
@@ -1,14 +1,31 @@
...
-  public event: StoreChangeEvent;
+export interface MutationRequest {
+  /**
+   * Type of event/operation
+   */
+  eventType: CRUDEvents;
+  /**
+   * Version used to detect changes in structure
+   */
+  version: number;
```
wtrocki, line 17: version for tracking changes of queue format

packages/offix-datastore/src/replication/mutations/MutationsQueue.ts
```diff
@@ -1,87 +1,169 @@
...
+      if (isOnline) {
+        const item = currentItems.shift();
+        logger("Mutation queue processing - online");
+        const operationPromise = this.options.client.mutation(item?.mutation as DocumentNode, item?.variables).toPromise();
+        operationPromise.then((data) => {
+          this.resultProcessor(currentItems, item as MutationRequest, data);
+        }).catch((error: CombinedError) => {
+          const retry = this.errorHandler(error, item as MutationRequest);
+          if (retry) {
+            this.items[0] = retry;
```
wtrocki, line 98: currentItems = Object.assign(this.items)

packages/offix-datastore/src/replication/queries/DeltaReplicator.ts
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  public async start() {
+    // If there is active pull available clear it
+    if (this.activePullInterval) {
+      clearInterval(this.activePullInterval);
+    }
+    // Only when online
+    if (await this.options.networkInterface.isOnline()) {
+      if (this.options.config.pullInterval) {
```
wtrocki, line 55: I forgot to add else here that should just do this.perform

packages/offix-datastore/src/replication/queries/DeltaReplicator.ts
```diff
@@ -0,0 +1,120 @@
...
+            this.perform();
+          }, this.options.config.pullInterval);
+        }).catch((error: CombinedError) => {
+          logger("Delta failed with error", error);
+          // We cannot handle errors on delta apart from notifying user
+          if (this.options.config.errorHandler) {
+            const result = this.options.config.errorHandler(error.networkError, error.graphQLErrors);
+            if (result) {
+              // Retry if user want it
+              this.perform();
```
wtrocki, line 67: No error catched for this one. It will fail on top level app

packages/offix-datastore/src/replication/queries/DeltaReplicator.ts
```diff
@@ -0,0 +1,120 @@
...
+          (async () => {
+            // TODO impove how we merge data together
+            const results = await model.update(d);
+            if (results.length === 0) {
+              // no update was made, save the data instead
+              model.save(d);
+              return;
+            }
+          })();
+        });
```
wtrocki, line 115: forgot to add this.performLock=false

packages/offix-datastore/src/replication/utils/delay.ts
```diff
@@ -0,0 +1,11 @@
+
+
+/**
+ * Delay execution of specific code by using promise await api
```
wtrocki, line 4: Not used. Experiment that can be deleted

packages/offix-datastore/src/storage/adapters/IndexedDBStorageAdapter.ts
```diff
@@ -10,151 +10,158 @@ const logger = createLogger("idb");
...
+  private transaction?: IDBTransaction;
+
+  constructor(transaction?: IDBTransaction) {
+    this.indexedDB = new Promise((resolve, reject) => {
+      this.resolveIDB = resolve;
+      this.rejectIDB = reject;
+    });
+    this.transaction = transaction;
+  }
+  // TODO Wrong architecture. Store can be created on demand
```
wtrocki, line 26: Not possible (yet) Unless we create new db every time

PR #656, opened 2020-07-24
json schema layer

kingsleyzissou:
##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-datastore/src/ModelSchema.ts
+++ b/packages/offix-datastore/src/ModelSchema.ts
@@ -101,3 +101,7 @@ function extractEncryptedFields<T = any>(schema: DSJsonSchema<T>): string[] {
   }
   return encrypted;
 }
+
+export function createModelSchema<T = any>(schema: DSJsonSchema<T>): ModelSchema<T> {
+  return new ModelSchema<T>(schema);
+}
--- a/packages/offix-datastore/src/ModelSchema.ts
+++ b/packages/offix-datastore/src/ModelSchema.ts
@@ -23,7 +23,8 @@ export declare class DSJsonSchema<T> {
 };
 
 export class ModelSchema<T = any>{
-  private schema: DSJsonSchema<T>;
+  private name: string;
+  private namespace: string;
   private primaryKey: string;
   private fields: string[];
   private encrypted: string[];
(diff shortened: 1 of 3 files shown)


kingsleyzissou:
This is still the early stages. The next step is to bring this into the datastore and create this layer before passing this to the model. I need to sort out some of the typings before I do that.

TODO:
- write some tests for this
- complete validation method or otherwise keep validation per field level as it currently is
- add the comments for the types and the methods
- hook this up to the rest of the datastore

I am just creating the PR for a bit of visibility and early feedback

@Eunovo @wtrocki

wtrocki:
Perfecto. This PR has to go first so I will deal with possible conflicts on my side

packages/offix-datastore/src/ModelSchema.ts
```diff
@@ -0,0 +1,129 @@
+import invariant from "tiny-invariant";
+import { JSONSchema7 } from "json-schema";
+
+export interface DSProperties extends JSONSchema7 {
```
wtrocki, line 4: Not particularly like those names

packages/offix-datastore/src/ModelSchema.ts
```diff
@@ -0,0 +1,129 @@
...
+  index?: boolean;
+  primary?: boolean;
+  default?: any;
+  encrypted?: boolean;
+  key?: string;
```
wtrocki, line 9: What is key?

packages/offix-datastore/src/ModelSchema.ts
```diff
@@ -0,0 +1,129 @@
...
+}
+
+export declare class DSJsonSchema<T> {
+  name: string;
+  namespace?: string;
+  version: number;
+  indices?: string[];
```
wtrocki, line 16: Indexes sound better?

packages/offix-datastore/src/ModelSchema.ts
```diff
@@ -0,0 +1,129 @@
...
+  encrypted?: string[];
+  primaryKey?: string;
+  type: "object";
```
wtrocki, line 19: Create type and add scalar as well

kingsleyzissou: @wtrocki type is there already, will add scalar

PR #657, opened 2020-07-24
Datasync cli

Eunovo:
### Description

Early stage of work on graphback plugin to generate datasync files. Currently;
- json schema for models are generated
- datasync config file initializing all the models is generared

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included
- [x] integration with sample app


packages/offix-datasync-codegen/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "name": "@offix/datasync-plugin",
+  "version": "0.1.0",
+  "description": "Graphback plugin for offix-datasync ",
```
wtrocki, line 4: Client

packages/offix-datasync-codegen/declarations.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'graphql-metadata';
```
wtrocki, line 1: Is there issue with typings for metadata?

packages/offix-datasync-codegen/src/OffixDataSyncConfig.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Config for OffixDataSync plugin
+ */
+export interface IOffixDataSyncPluginConfig {
+    /**
+     * The output dir path.
+     * The json schema will be written to ${outputDir}/schema.json
+     * The datasync config will be written to ${outputDir}/config.ts
+     */
+    outputDir?: string;
```
wtrocki, line 10: Rename to model output dir

packages/offix-datasync-codegen/tests/__snapshots__/DataSyncClientPlugin.test.ts.snap
```diff
@@ -0,0 +1,71 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Test datasync config 1`] = `
```
wtrocki, line 3: Not sure if we need to generate config or it should be just part of the docs.

packages/offix-datasync-codegen/declarations.d.ts
Eunovo: Yes

wtrocki:
Amazing work. 

## Crazy idea for UX

command should ask couple questions - what is your desired output config etc and then create GraphQL.config file. For that we will need to extend graphback cli :( so it might be too much for this PR but it might be worth to chat about it

packages/offix-datasync-codegen/tests/__snapshots__/DataSyncClientPlugin.test.ts.snap
Eunovo: We can generate but still leave it editable to the user. Like generate boilerplate code. To ensure that we don't overwrite the user's work, we can add a config option to the plugin to skip generating datasync config

packages/offix-datasync-codegen/declarations.d.ts
wtrocki: Cc @craicoverflow Let's fix them then

packages/offix-datasync-codegen/tests/__snapshots__/DataSyncClientPlugin.test.ts.snap
wtrocki: Trick is to not generate anything that should be editable, but ok to leave it

Eunovo:
@wtrocki  @kingsleyzissou I have added graphback to sample app and some cli documentation. The docs are far behind and need explanation on the schema fields but this PR should be mergeable. Also, I couldn't add much to the schema.json because I'm not sure how most of the new fields should work but It should be easy to edit what is generated.
Please review and merge is no changes are required.

Eunovo:
@wtrocki It won't be easy to test the cli integration without publishing the plugin. I'm experiencing problems using the linked packages on my end due to problems related to multiple installed graphql modules.

kingsleyzissou:
@Eunovo will review now. As for the docs, I think things are changing quite quickly, so I would say maybe we can nail down the architecture and our MVPs and focus on the docs when we have something a bit more solid. @wtrocki WDYT?

wtrocki:
> @wtrocki WDYT?

See https://github.com/aerogear/offix/issues/659

@kingsleyzissou let's publish package for eunovo and add myself as maintainer (aerogear-ci) as well

docs/ref-datastore-getting-started.md
```diff
@@ -101,3 +101,36 @@ The `DataStore` constructor takes a schema version parameter(defaults to 1).
...
+You can import models from `config.ts` and start coding! 
+
+### Installing the plugin
+
+For npm
+`npm install offix-datasync-client-plugin`
+or yarn
+`yarn add offix-datasync-client-plugin`
+
+Add the plugin to your `.graphqlrc.yml`
```
wtrocki, line 120: you will not have .graphqlrc.yml. That is why we need small wrapper that will create it with some sensible defaults for our plugin

wtrocki: This can be separate issue

examples/react-datastore/.graphqlrc.yml
```diff
@@ -0,0 +1,8 @@
+schema: './src/schema.graphql'
+extensions:
+  graphback:
+    # path to data mode file(s)
+    model: './src/model/runtime.graphql'
+    plugins:
+      offix-datasync-client-plugin:
```
wtrocki, line 7: I think it is best for us to also build schema for developers to see/share etc.

Eunovo: Can we take out the schema reference? I tried removing it and got some errors

wtrocki: This goes against this file
This file is build to describe schema. 
You can get schema from link if you wish.... 

But generally we should generate visible schema for users - drop schema plugin explicitly and dump schema somewhere.

PR #671, opened 2020-07-28
Integrate sample app

Eunovo:
### Description

I have fixed a couple of problems hindering integration here. I still need the package to be published before I can test new changes. @kingsleyzissou  I also added primary key. I also generate a schema export file to combat the issue of types. The schema export file exports the schema with proper types to avoid type errors.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works


packages/offix-datastore/src/ModelSchema.ts
```diff
@@ -11,6 +11,10 @@ export interface DataSyncProperties extends JSONSchema7 {
   primary?: boolean;
   default?: any;
   encrypted?: boolean;
+  /**
```
kingsleyzissou, line 14: @Eunovo thanks, yeah I cut a few corners to get the last PR merged. I need to come back and add comments to everything.

Eunovo: The `key` param carried over from my code though. So, no problem

packages/offix-datasync-codegen/tests/__snapshots__/DataSyncClientPlugin.test.ts.snap
```diff
@@ -2,74 +2,95 @@
...
   },
-]
+  "schemaExport": "import { DataSyncJsonSchema } from \\"offix-datastore\\";
+import schema from \\"./schema.json\\";
+
+type Schema<T = any> = {
+    [P in keyof T]: DataSyncJsonSchema<any>
+};
+
+export schema as Schema;
```
kingsleyzissou, line 93: I think this might need to be `export default schema ...`

Eunovo: Yes. I'm changing this currently

kingsleyzissou: @Eunovo cool your solution was better because we get a named export.

Eunovo: I needed to explicitly set the type of the export and I didn't know how to do that in default export

examples/react-datastore/src/datasync/schema.json
```diff
@@ -0,0 +1,22 @@
+{
```
kingsleyzissou: @Eunovo think this file can go now. We have the `schema` subdirectory with the `schema.json` and `index.ts`

kingsleyzissou: Nevermind, you've done that already

PR #677, opened 2020-07-29
Query optimization

Eunovo:
### Description

This PR is intended to contain work to optimize predicate execution in datasync by using indexes. Approach has been oultined in this draft with comments about how the algorithm should work. 

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/offix-datastore/src/storage/adapters/indexeddb/PredicateExecutor.ts
+++ b/packages/offix-datastore/src/storage/adapters/indexeddb/PredicateExecutor.ts
@@ -1,0 +1,71 @@
+import { PredicateFunction, PredicateExpression, ModelFieldPredicate } from "../../../predicates";
+
+const getIndexedOperations = (predicate: PredicateFunction): any[] => {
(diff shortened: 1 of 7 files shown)


packages/offix-datastore/src/storage/adapters/indexeddb/PredicateExecutor.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    return []; // return all
+}
+
+const handleAndExpression = (predicates: PredicateFunction[]) => {
+    let data: any[] = [];
+    // TODO
+    // To reduce search space here
+    // it seems reasonable to use most constraining index
+    // and filter the data returned by most constraining index
```
wtrocki, line 32: https://developer.mozilla.org/en-US/docs/Web/API/IDBKeyRange check options here for reference. SQL lite allows much more including joins etc.

wtrocki:
Closing as outdated.

PR #687, opened 2020-07-31
React hooks

Eunovo:
### Description

This PR implements react hooks for offix-datasync. Some strategies for fetching and updating data will be implemented as soon as they are implemented in Models. The hooks are also integrated in the sample app. I haven't verified the hooks because verification with the linked local package is blocked by react. I recommend that the new changes be published before the hooks can be used.

##### Checklist

- [x] `npm run build` works

### DIFF
--- a/examples/react-datastore/src/App.tsx
+++ b/examples/react-datastore/src/App.tsx
@@ -1,4 +1,5 @@
 import React, { useState, useEffect } from 'react';
+import { CRUDEvents } from 'offix-datastore';
 import { Button } from 'antd';
 import 'antd/dist/antd.css';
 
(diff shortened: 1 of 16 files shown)


examples/react-datastore/src/App.tsx
```diff
@@ -9,7 +10,14 @@ function App() {
 
   const [mounted, setMounted] = useState<boolean>(false);
   const [addView, setAddView] = useState<boolean>(false);
-  const  { loading, error, data } = useFindTodos();
+  const  { isLoading: loading, error, data, subscribeToMore } = useFindTodos();
+  useEffect(() => {
```
wtrocki, line 14: Why use effect is needed? Apollo doesn't have it

Eunovo: See https://www.apollographql.com/docs/react/data/subscriptions/#subscribing-to-updates-for-a-query

Calling subscribeToMore in a `componentDidMount` is equivalent to calling it in a `useEffect`. The user can also provide a clean-up function to unsubscribe. In a class based component, you do the clean up in a `componentWillUnmount`

PR #691, opened 2020-08-04
Move events to model

Eunovo:
### Description

In this PR, I have moved events to Model. We need to implement update strategies so replication is no longer event-based. Now we can fire events after replication or before replication. We need to rethink the information in change events since replication no longer uses events. I didn't change the code too much, there are still lots of improvements we can make now that replication is called in Model.

### DIFF
--- a/packages/offix-datastore/src/DataStore.ts
+++ b/packages/offix-datastore/src/DataStore.ts
@@ -6,6 +6,7 @@ import { IndexedDBStorageAdapter } from "./storage/adapters/IndexedDBStorageAdap
 import { ModelSchema, DataSyncJsonSchema } from "./ModelSchema";
 import { DataStoreConfig } from "./DataStoreConfig";
 import { createLogger, enableLogger } from "./utils/logger";
+import { ModelReplicationConfig } from "./replication/api/ReplicationConfig";
 
 const logger = createLogger("DataStore");
 // TODO disable logging before release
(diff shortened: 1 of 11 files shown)


wtrocki:
Looking at this

packages/offix-datastore/src/DataStore.ts
```diff
@@ -35,30 +35,36 @@ export interface CustomEngines {
...
-    this.engines = engines;
 
     if (engines && engines.storeAdapter) {
       this.storage = new LocalStorage(engines.storeAdapter);
     } else {
       const indexedDB = new IndexedDBStorageAdapter();
       this.storage = new LocalStorage(indexedDB);
     }
+
+    if (this.config.replicationConfig) {
```
wtrocki, line 53: -1 for that. 

There are async methods there. 
I think we might need to have init with await and we cannot have this code in constructor.

packages/offix-datastore/src/DataStore.ts
```diff
@@ -35,30 +35,36 @@ export interface CustomEngines {
...
+      logger("Replication configuration was not provided. Replication will be disabled");
+    }
   }
 
   public createModel<T>(schema: DataSyncJsonSchema<T>) {
     const modelSchema = new ModelSchema(schema);
     const model = new Model<T>(modelSchema, this.storage);
-    // TODO replace any
-    this.models.push(model as any);
+    this.replicator?.startModelReplication(model, this.storage);
```
wtrocki, line 67: So if we do this then we simply do not need replication init method. 
We can pass replication info to create model function..

packages/offix-datastore/src/Model.ts
```diff
@@ -72,8 +75,16 @@ export class Model<T = unknown> {
...
-  public save(input: T): Promise<T> {
-    return this.storage.save(this.schema.getStoreName(), input);
+  public async save(input: T): Promise<T> {
+    const data = await this.storage.save(this.schema.getStoreName(), input);
+    this.replicator?.replicate(data, CRUDEvents.ADD);
+    this.changeEventStream.publish({
+      eventType: CRUDEvents.ADD,
+      data,
+      storeName: this.getStoreName(),
+      eventSource: "user"
```
wtrocki, line 85: This is no longer needed.

PR #692, opened 2020-08-04
chore: add sqlite adapter

kingsleyzissou:
### Description
Work in progress for sql adapter. For now I am using websql as it is easier to work with locally.
Once that is running fine locally, I will implement Native support


packages/offix-datastore/src/replication/api/MetadataModels.ts
```diff
@@ -0,0 +1,30 @@
+import { ModelSchema } from "../../ModelSchema";
+import { MODEL_METADATA, MODEL_METADATA_KEY, MUTATION_QUEUE, MUTATION_QUEUE_KEY } from "./Replicator";
+
+export const metadataModel = new ModelSchema<any>({
```
wtrocki, line 4: Love it! Strongly typed metadata. It will make sense to merge this as.. 

@Eunovo work will invalidate this changes - we need to change id etc.

packages/offix-datastore/src/storage/adapters/IndexedDBStorageAdapter.ts
```diff
@@ -11,26 +11,30 @@ const logger = createLogger("idb");
...
+  constructor(dbName: string, schemaVersion: number, transaction?: IDBTransaction) {
+    this.dbName = dbName;
+    this.schemaVersion = schemaVersion;
     this.indexedDB = new Promise((resolve, reject) => {
       this.resolveIDB = resolve;
       this.rejectIDB = reject;
     });
     this.transaction = transaction;
   }
   // TODO Wrong architecture. Store can be created on demand
```
wtrocki, line 30: This TODO can be removed :D

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
```diff
@@ -0,0 +1,250 @@
...
+        );
+      });
+    });
+  }
+}
+
+// TODO change to constants/enums
+const prepareStatement = (input: any, type: string = "insert"): [string, any[]] => {
+  if (type === "insert") {
+    input.id = generateId();
```
wtrocki, line 218: Love it!

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
```diff
@@ -0,0 +1,250 @@
...
+    const query = `DELETE FROM ${storeName} ${condition}`;
+    // @ts-ignore
+    return this.transaction(query, [predicate.value]);
+  }
+
+  public getSQLiteInstance() {
+    return this.sqlite;
+  }
+
+  createTransaction(): Promise<StorageAdapter> {
```
wtrocki, line 97: Let's crete separate issue. We can return this for the moment

packages/offix-datastore/package.json
```diff
@@ -52,6 +53,7 @@
     "tiny-invariant": "1.1.0",
     "urql": "1.9.8",
     "uuid": "8.3.0",
+    "websql": "^1.0.0",
```
wtrocki, line 56: So library says that this is only for the node.js which will increase size of the package. 

https://www.npmjs.com/package/websql

Alternatives we can play with 
https://github.com/gdbate/lib-websql

https://www.npmjs.com/package/@types/sqlite3 (and sqllite3 as peerDependency)

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
```diff
@@ -0,0 +1,250 @@
...
+
+  public async save(storeName: string, input: any): Promise<any> {
+    const [cols, vals] = prepareStatement(input, "insert");
+    const stmt = `INSERT INTO ${storeName} ${cols}`;
+    return this.transaction(stmt, vals);
+  }
+
+  public async query(storeName: string, predicate?: PredicateFunction): Promise<any> {
+    if (!predicate) {
+      return await this.getStore(storeName);
```
wtrocki, line 69: I will rename it to `fetchAll`

packages/offix-datastore/package.json
kingsleyzissou: @wtrocki Oh okay cool, will have a look at this, this might be what I need.

I sorted out the types stuff (I forgot to add them to to `tsconfig`)

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
kingsleyzissou: @wtrocki I think this is what it might be called in the interface, but I can rename it and update in the `indexeddb` adapter too

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
kingsleyzissou: @wtrocki I'll just leave them unimplemented for now then - if they are not there we get typescript errors all over the place.

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
```diff
@@ -0,0 +1,250 @@
...
+    }
+    const condition = predicateToSQL(predicate as ModelFieldPredicate);
+    const query = `SELECT * FROM ${storeName} ${condition}`;
+    // @ts-ignore
+    const res = await this.readTransaction(query, [predicate.value]);
+    return res;
+  }
+
+  public async update(storeName: string, input: any, predicate?: PredicateFunction): Promise<any> {
+    const condition = predicateToSQL(predicate as ModelFieldPredicate);
```
kingsleyzissou, line 79: @Eunovo I actually have a question for this, when is the `ModelPredicate` used and when is the normal `Predicate` function used. With the SQL adapter, I kind of need to guarantee that I am getting the model predicate in order to perform some of the queries.

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
kingsleyzissou: Nope, I was wrong. Changing it now

packages/offix-datastore/src/replication/api/MetadataModels.ts
kingsleyzissou: @wtrocki yeah this was needed because we need to know the field names for all the stores upfront so we could create those, so I needed the full schema for the models and it just made sense to convert these to model schema too

packages/offix-datastore/src/storage/adapters/WebSQLAdapter.ts
wtrocki: Yep.. We can still do in memory filtering on layer above.

PR #701, opened 2020-08-07
Sqlite refactor

kingsleyzissou:
### Description

Just a small refactor. This now works in the sample app (except for the update method).

To verify add the following to the `datastore` config in the `datasync/config.ts` file:

```
export const datastore = new DataStore({
  ...
}, {
  storeAdapter: new WebSQLAdapter("offixdb", 2)
});

```

### DIFF
--- a/packages/datastore/datastore/src/storage/adapters/WebSQLAdapter.ts
+++ b/packages/datastore/datastore/src/storage/adapters/WebSQLAdapter.ts
@@ -31,30 +31,23 @@ export class WebSQLAdapter implements StorageAdapter {
       "Offix datastore", 5 * 1024 * 1024
     );
   }
-  // TODO Wrong architecture. Store can be created on demand
+
   public addStore(config: ModelSchema) {
     this.stores.push(config);
   }
 
   public async createStores() {
     const existingStoreNames = await this.getStoreNames();
-    existingStoreNames.forEach((storeName) => {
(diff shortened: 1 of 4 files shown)


packages/datastore/datastore/src/storage/adapters/WebSQLAdapter.ts
```diff
@@ -110,35 +98,19 @@ export class WebSQLAdapter implements StorageAdapter {
...
+    return res.map((r) => r.name);
   }
 
-  private async transaction(query: string, args: any) {
+  private async transaction(query: string, args: any): Promise<any[]> {
     return new Promise((resolve, reject) => {
       const successCb = (tx: SQLTransaction, res: SQLResultSet) => {
         logger("Executed query with results", res.rows);
-        resolve(res.rows);
+        resolve(flattenResultSet(res.rows));
```
wtrocki, line 113: Why this is needed?

wtrocki:
Verified. Worth to merge as @Eunovo will need to rebase this

packages/datastore/datastore/src/storage/adapters/WebSQLAdapter.ts
kingsleyzissou: @wtrocki this was needed because the results came back as an SQLResult set object rather than an array of results. So I was getting errors when trying to use this.

PR #702, opened 2020-08-07
Better filters

Eunovo:
### Description

Implement Object-based filters for IndexedDB and add base for WebSQL. Also, added feature to infer Filter type from Model. See the attached image to verify filter types.

Screenshot (37)

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


Eunovo:
I'm going to rebase and drop old commits

Eunovo:
I'm still working on this. I need to add some comments to code

packages/datastore/datastore/src/replication/queries/DeltaReplicator.ts
```diff
@@ -2,7 +2,7 @@ import { ModelDeltaConfig } from "../api/ReplicationConfig";
 import { NetworkStatusEvent, NetworkStatus } from "../../network/NetworkStatus";
 import { DocumentNode } from "graphql";
 import { Client, OperationResult, CombinedError } from "urql";
-import { convertPredicateToFilter } from "../utils/convertPredicateToFilter";
+import { convertFilterToGQLFilter } from "../utils/convertFilterToGQLFilter";
```
wtrocki, line 5: Thank you for doing this!

packages/datastore/datastore/src/storage/adapters/WebSQLAdapter.ts
```diff
@@ -231,12 +231,25 @@ const prepareStatement = (input: any, type: string = "insert"): [string, any[]]
...
+    }
+    const op = Object.keys(filter[key])[0];
+    const operator = (op === "eq") ? "=" : undefined;
+    invariant(operator, "Operator not supported");
+
+    const value = filter[key][op];
+    tokens.push(`${key}=${value}`);
+  });
+
+  return `WHERE ${tokens.join(" AND ")} ?`;
```
wtrocki, line 252: I guess we do not care about SQL Injection here?

wtrocki:
Do we have validation for invalid filter?
AM I right to assume that there is only single way to express all filtering and it is `Filter`.
What happened to ModelFilter and other stuff?

packages/datastore/datastore/src/storage/adapters/WebSQLAdapter.ts
Eunovo: I only dropped base work for SQL support

packages/datastore/datastore/src/react/hooks/delete.ts
```diff
@@ -1,17 +1,17 @@
 import { useReducer } from "react";
 import { Model } from "../../Model";
 import { reducer, InitialState, ActionType } from "../ReducerUtils";
-import { Predicate } from "../../predicates";
+import { Filter } from "../../filters";
```
wtrocki, line 4: Name is too generic for me. I will add something like StoreFiter or something

Eunovo:
@wtrocki  We can merge this PR. There are some failing tests related to transactions that I can investigate in a separate PR. There's also the matter of sample app integration that I will handle in a separate PR.

wtrocki:
Let's fix test and I'm happy to get that to master for more testing locally.

wtrocki:
Our master will be broken now.

PR #708, opened 2020-08-10
Filter to sql

Eunovo:
### Description

Implemented a converter for filters to SQL expressions. Tests are include for verification,

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/datastore/datastore/src/storage/adapters/websql/filterToSQL.ts
+++ b/packages/datastore/datastore/src/storage/adapters/websql/filterToSQL.ts
@@ -35,7 +35,7 @@ export const filterToSQL = (filter?: Filter) => {
 
     Object.keys(filter).forEach((key) => {
         if (!(filter[key] instanceof Object)) {
-            tokens.push(`eq=${filter[key]}`);
+            tokens.push(OperatorToSQLMap.eq(key, filter[key]));
             return;
         }
         const op = Object.keys(filter[key])[0];
--- a/packages/datastore/datastore/src/storage/adapters/websql/filterToSQL.ts
+++ b/packages/datastore/datastore/src/storage/adapters/websql/filterToSQL.ts
@@ -1,0 +1,50 @@
+import invariant from "tiny-invariant";
+import { AllOperators, Filter } from "../../..";
+
+type OperatorToSQL = {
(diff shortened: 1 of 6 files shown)


wtrocki:
Will take look early tommorow

packages/datastore/datastore/src/storage/adapters/websql/filterToSQL.ts
```diff
@@ -0,0 +1,70 @@
...
+const defaultOp = (op: string) => (
+    (key: string, value: any) => {
+        if ("string" === typeof value) {
+            value = `'${value}'`;
+        }
+        return `${key} ${op} ${value}`;
+    }
+);
+
+// TODO contains is tricky, verify solution
```
wtrocki, line 17: Was this verified.

packages/datastore/datastore/src/storage/adapters/websql/utils.ts
```diff
@@ -0,0 +1,46 @@
+import invariant from "tiny-invariant";
+import { generateId } from "../..";
+
+// TODO change to constants/enums
+export const prepareStatement = (input: any, type: string = "insert"): [string, any[]] => {
+    if (type === "insert") {
+        input.id = generateId();
```
wtrocki, line 7: No the best place to generate id.

wtrocki:
Generating SQL statements seems like very well known and understood problem with handful of the packages and solutions we can use that are mature. Any reason why we went for custom code?

While it is ok to have it written, we need to think about maintenance of that and possible bugs.

wtrocki:
Played with unit tests. Good for merge! Please do follow up improvements as separate PR.

packages/datastore/datastore/src/storage/adapters/websql/filterToSQL.ts
Eunovo: I need to run this to verify it. We still have problem of running sample app

Eunovo:
> Generating SQL statements seems like very well >known and understood problem with handful of the >packages and solutions we can use that are mature. > Any reason why we went for custom code?

TBH, I didn't check for existing packages. I didn't think there would be any tailored to our filters

Issue #712, opened 2020-08-11
SQLite does not support array types


Eunovo:
While testing filtering with `contains` operation, I discovered that array types are not saved to the db when using the websql adapter. After some digging, I realised that there is no support for an "Array" type so we can't save arrays naturally. 

I think we have two options here,
- Don't allow array types when using websql
- Use some alternate method to save arrays, there are a couple of ways to this, see [this](https://stackoverflow.com/questions/3005231/how-to-store-array-in-one-column-in-sqlite3)

@wtrocki @kingsleyzissou

wtrocki:
Split will work only for single dimensional arrays. 
I have been playing with SQL lite and I think that we do not need to create proper tables. 
Our data can be saved as one column 

So our API can be:

```
await AsyncStorage.setItem( "mykey", "myValue")
```

However indexed fields will still need to be brought into columns for searchability.

PR #713, opened 2020-08-11
Quick fixes and IDB Cursor Support

Eunovo:
### Description

I modified the sample app to handle change events. I also tried to verify that the `contains` op works when using websql. It works for strings but not for arrays (I created test cases for both). I created an issue detailing the problem with arrays on websql #712. Getting the sample to work with the linked package was tricky but you can run it if you go into `offix/node_modules` and delete the react folder. Note that "react" will be re-installed the next time you run `yarn install` inside the workspace whether there's a react dependency in datastore or not because "react-offix-hooks" depends on it too. I also removed the react dependency but had to add it again because "urql" requires it and the tests won't run. Other small fixes are also in this PR.

#### EDIT

I have added cursor support for IDBAdapter to this PR

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/examples/react-datastore/src/App.tsx
+++ b/examples/react-datastore/src/App.tsx
@@ -6,18 +6,41 @@ import 'antd/dist/antd.css';
 import { useFindTodos } from './helpers/hooks';
 import { TodoList, AddTodo, Loading, Error, Header } from './components';
 
+const onTodoAdded = (currentData: any[], newData: any) => {
+  if (!currentData) return [newData];
+      return [...currentData, newData];
+}
+
+const onTodoChanged = (currentData: any[], newData: any[]) => {
+  if (!currentData) return [];
+
+  return currentData.map((d) => {
+    const index = newData.findIndex((newD) => newD.id === d.id);
+    if (index === -1) return d;
+    return newData[index];
+  });
+}
+
+const onTodoRemoved = (currentData: any[], removedData: any[]) => {
+  if (!currentData) return [];
+  return currentData
+    .filter(
+      (d) => removedData.findIndex((newD) => newD.id === d.id)
+    );
+}
+
 function App() {
 
-  const [mounted, setMounted] = useState<boolean>(false);
   const [addView, setAddView] = useState<boolean>(false);
   const  { isLoading: loading, error, data, subscribeToMore } = useFindTodos();
   useEffect(() => {
-    const subscription = subscribeToMore(CRUDEvents.ADD, (newData) => {
-      if (!data) return [newData];
-      return [...data, newData];
-    });
-    return () => subscription.unsubscribe();
-  }, []);
+    const subscriptions = [
+      subscribeToMore(CRUDEvents.ADD, (newData) => onTodoAdded(data, newData)),
+      subscribeToMore(CRUDEvents.UPDATE, (newData) => onTodoChanged(data, newData)),
+      subscribeToMore(CRUDEvents.DELETE, (newData) => onTodoRemoved(data, newData)),
+    ];
+    return () => subscriptions.forEach(s => s.unsubscribe());
+  }, [data, subscribeToMore]);
 
   if (loading) return <Loading />;
 
--- a/packages/datastore/datastore/src/react/hooks/query.ts
+++ b/packages/datastore/datastore/src/react/hooks/query.ts
@@ -9,6 +9,10 @@ const createSubscribeToMore = (model: Model, dispatch: Dispatch<Action>) => {
         // TODO subscribe to specific predicate
         const subscription = model.subscribe(eventType, (event) => {
             const newData = updateResult(event.data);
+            if (subscription.closed) {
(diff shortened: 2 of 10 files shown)


packages/datastore/datastore/src/react/hooks/query.ts
```diff
@@ -9,6 +9,10 @@ const createSubscribeToMore = (model: Model, dispatch: Dispatch<Action>) => {
         // TODO subscribe to specific predicate
         const subscription = model.subscribe(eventType, (event) => {
             const newData = updateResult(event.data);
+            if (subscription.closed) {
+                // Important to check beacuse Componnent could be unmounted
```
wtrocki, line 13: But this means that we have memory leaks and we hiding the error that is causing them?

packages/datastore/datastore/src/storage/adapters/indexedDB/IndexedDBStorageAdapter.ts
```diff
@@ -115,13 +115,31 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
...
     const store = await this.getStore(storeName);
-    const all = await this.convertToPromise<any[]>(store.getAll());
+    const cursorReq = store.openCursor();
+    const result: any[] = [];
+
+    return new Promise((resolve, reject) => {
+      cursorReq.onsuccess = () => {
+        const cursor = cursorReq.result;
+        if (cursor) {
+          if (!filter) {
```
wtrocki, line 127: IMHO if there is no filter we should not use cursor.

packages/datastore/datastore/src/storage/adapters/indexedDB/IndexedDBStorageAdapter.ts
```diff
@@ -115,13 +115,31 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
...
+            result.push(cursor.value);
+          } else {
+            const predicate = getPredicate(filter);
+            if (predicate(cursor.value)) {
```
wtrocki, line 131: Why we have done this? Any reason to use cursor vs full data read?

wtrocki: Any issue where we discussed that? Want to get more context

packages/datastore/datastore/src/storage/adapters/websql/WebSQLAdapter.ts
```diff
@@ -47,19 +48,21 @@ export class WebSQLAdapter implements StorageAdapter {
   }
 
   public async save(storeName: string, input: any): Promise<any> {
+    input.id = generateId();
```
wtrocki, line 51: Perfect place for id! Thx. I will follow up with this.

packages/datastore/datastore/src/react/hooks/query.ts
Eunovo: I don't think so. I added this check incase this listener is active when the subscription was ended. The listener won't be called for new events but what the listener that was just called before unsubscription

examples/react-datastore/src/App.tsx
```diff
@@ -6,18 +6,41 @@ import 'antd/dist/antd.css';
 import { useFindTodos } from './helpers/hooks';
 import { TodoList, AddTodo, Loading, Error, Header } from './components';
 
+const onTodoAdded = (currentData: any[], newData: any) => {
+  if (!currentData) return [newData];
+      return [...currentData, newData];
+}
+
+const onTodoChanged = (currentData: any[], newData: any[]) => {
```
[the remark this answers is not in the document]
Eunovo: The code here only controls user state not the DB.

packages/datastore/datastore/src/storage/adapters/indexedDB/IndexedDBStorageAdapter.ts
Eunovo: I don't think we discussed this in an issue be we have discussed this on the discord chat. Reason for Cursors is to avoid loading all data into memory. Without Cursors, we have to load all data into memory before filtering. Although, If there is no filter, we don't need Cursors since we still load all data. But when there is a filter, we only keep results in memory.

wtrocki: I do not like us adding some optimatizations like results in memory. Performance etc. Without actually having performance tests and general idea how this works. Let's keep it but please do not do any optimalizations without discussion

Eunovo:
@wtrocki  Conflicts have been resolved

examples/react-datastore/src/App.tsx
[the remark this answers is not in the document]
wtrocki: Like this looks like deduplication issue that should be happening in core.

PR #717, opened 2020-08-12
Global replication

wtrocki:
Replication engine version 600 :D. 
This has still many TODO's and probably not functional as whole. 
What i need to do is to start writing integration/unit tests for this next week.

I'm going to be on the PTO tomorrow and Friday so feel free to merge this to master or ignore it - depending on reviews. Sorry for such long process of making changes.
I believe that with this architecture things will speed up nicely. 

Kudos to @kingsleyzissou for amazing changes on schema. Having abstraction for ID helps in many places here, but I still feel that our model save/update methods need more work.

Kudos for building nice translation engine for this @Eunovo 

Progress:

- [x] Mutation replication
- [x] Delta replication
- [ ] Subscriptions
- [ ] Unit tests
- [ ] Storage improvements
- [ ] Error handling and queue updates


wtrocki:
Fixing and testing sample app. Having problems with duplicate graphql instances.

wtrocki:
Getting error in app:

```
×
Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
```

@Eunovo do you seen that before?

Eunovo:
@wtrocki this is the problem with multiple react modules in one project that I have been referring to. To run the sample app with the linked package, you must manually delete react from `offix/node_modules`. Note that every time you run `yarn install` react will be reinstalled because several packages depend on it. So, you have to Manually delete the react module every time or you can't run sample app with linked Datastore

wtrocki:
Maybe we can add this workaround to package.json scripts. It took me some time to figure it out. If you see any problems and know solutions please share them with team.

packages/datastore/datastore/src/replication/mutations/MutationsQueue.ts
```diff
@@ -57,124 +108,155 @@ export class MutationsReplicationQueue {
...
+    const storeName = model.getStoreName();
+    const operations = this.modelMap[storeName];
+    invariant(operations, "Missing GraphQL mutations for replication");
+
+    const mutationRequest = {
+      storeName,
+      data,
+      eventType
+    };
+    await this.enqueueRequest(mutationRequest);
```
Eunovo, line 142: It seems like provision has made here to include the enqueue operation in the CRUD transaction. But the passed store object is not actually used. If this is a TODO or a mistake, I can fix it for you

Eunovo:
> model save/update methods need more work.

@wtrocki  What do you have in mind? I can make the improvements

wtrocki:
I think similar to pouch we need to have saveOrUpdate method. As you see also deleting by ID is kinda ackward now. 

Also I think I have broken tests for save/update etc. We now await for replcation request to be saved

packages/datastore/datastore/src/replication/mutations/MutationsQueue.ts
wtrocki: mistake. This might fix tests

Eunovo:
I'm merging this

Eunovo:
> I think similar to pouch we need to have saveOrUpdate method.

Any advantages?

Eunovo:
> As you see also deleting by ID is kinda ackward now.

Makes sense to have updateById and deleteById so users can take advantage of primary key indexing to improve performance. I'll add these methods in a separate PR

wtrocki:
Going back to this tommorow

PR #722, opened 2020-08-14
Query, Update and Delete by ID

Eunovo:
### Description

Implementation of Query, Update and Delete by ID. Here, I added three new methods to our API that performs, query, update and delete operations using the ID. Although the methods have "ById" in their names, I do not use "Hardcoded id", I obtain the primary key from the store schema and use it for the operations. It might be better for us to by Primary Key(like updateByPrimaryKey) instead since we are not actually using "id". 

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/datastore/datastore/src/Model.ts
+++ b/packages/datastore/datastore/src/Model.ts
@@ -83,7 +83,11 @@ export class Model<T = unknown> {
     return this.storage.query(this.schema.getStoreName(), filter);
   }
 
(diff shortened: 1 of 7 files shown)


wtrocki:
Awesome! This will invalidate some helpers I have build

wtrocki:
We also need save or update by ID

PR #725, opened 2020-08-17
Datastore save or update

Eunovo:
### Description

Implementation of `saveOrUpdate` for both IDB and WebSQL

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/datastore/datastore/src/storage/adapters/indexedDB/IndexedDBStorageAdapter.ts
+++ b/packages/datastore/datastore/src/storage/adapters/indexedDB/IndexedDBStorageAdapter.ts
@@ -110,10 +110,10 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
 
   public async save(storeName: string, input: any) {
     const store = await this.getStore(storeName);
-    // TODO hardcoded id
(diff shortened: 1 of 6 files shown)


packages/datastore/datastore/src/storage/adapters/indexedDB/IndexedDBStorageAdapter.ts
```diff
@@ -162,9 +162,34 @@ export class IndexedDBStorageAdapter implements StorageAdapter {
 
   public async updateById(storeName: string, input: any, id: string) {
     const store = await this.getStore(storeName);
-    const primaryKey = getPrimaryKey(this.stores, storeName); // TODO keypath could be an array
-    await this.convertToPromise<IDBValidKey>(store.put({ ...input, [primaryKey]: id }));
-    return this.convertToPromise(store.get(id));
+    const primaryKey = getPrimaryKey(this.stores, storeName);
```
wtrocki, line 165: I think `getPrimaryKey` should be part of the model. Also see getPrimaryKeyFilter that I have added to model. 

I think I will refactor this after we merge.

wtrocki:
how this affect our hooks. DO we need extra issue to cover this?

PR #741, opened 2020-08-18
New datastore cli

Eunovo:
### Description

Use GraphbackPluginEngine(not GraphbackGenerator as previously suggested) to generate code so users are not forced to install graphback

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


Eunovo:
@wtrocki  We can't use GraphbackGenerator because it tries to load plugins as modules that are defined in package.json

Eunovo:
I added https://github.com/aerogear/offix/blob/new-datastore-cli/packages/datastore/datastore-codegen/tests/Cli.test.ts
to help with verification but its failing on CircleCI

wtrocki:
Lets make build pass and merge. There are some things missing but we need to work on rebased branch

PR #751, opened 2020-08-19
Add comments to model schema (fixes #740)

### LINKED ISSUE #740: Add comments to code
## Feature Request
Some of the features were merged pretty quickly and there was no time to add the comments. (ModelSchema and WebSQL adapter). I need to go back through the code and add adequate comments.

kingsleyzissou:
### Description

I have just gone in and added comments to the ModelSchema for future reference.

### DIFF
--- a/packages/datastore/datastore/src/ModelSchema.ts
+++ b/packages/datastore/datastore/src/ModelSchema.ts
@@ -14,11 +14,13 @@ export interface GenetratedModelSchema {
 }
 
 export interface ModelSchemaProperties extends JSONSchema7 {
-  /** Index field, whether or not the
+  /**
+   * Index field, whether or not the
    * field should be indexed
    */
   index?: boolean;
-  /** Primary key field flag if the field is
+  /**
+   * Primary key field flag if the field is
    * the primary key
    */
   primary?: boolean;
@@ -26,8 +28,8 @@ export interface ModelSchemaProperties extends JSONSchema7 {
    * Default value for the field
    */
   default?: any;
-  /** Flag for if the field should be encrypted
-   * i.e. password field
+  /**
+   * Flag for if the field should be encrypted
    */
   encrypted?: boolean;
   /**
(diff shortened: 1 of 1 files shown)


Eunovo (approved):

packages/datastore/datastore/src/ModelSchema.ts
```diff
@@ -14,28 +14,75 @@ export interface GenetratedModelSchema {
...
+  /** Primary key field flag if the field is
+   * the primary key
+   */
   primary?: boolean;
+  /**
+   * Default value for the field
+   */
   default?: any;
+  /** Flag for if the field should be encrypted
+   * i.e. password field
```
wtrocki, line 30:
```suggestion
   * i.e. password field
```

Password field should never be passed here :D

packages/datastore/datastore/src/ModelSchema.ts
```diff
@@ -56,52 +103,100 @@ export class ModelSchema<T = any>{
     this.encrypted = extractEncryptedFields(this.fields, schema.encrypted);
   }
 
+  /**
+   * Method to fill the object with default values
+   *
+   */
   public fill(): void {
```
wtrocki, line 110: It is redundant to have this method then? :)

packages/datastore/datastore/src/ModelSchema.ts
kingsleyzissou: @wtrocki so the value will never be passed there, but surely at some point the model will have to handle sensitive fields that might need encryption, or not really?

packages/datastore/datastore/src/ModelSchema.ts
kingsleyzissou: @wtrocki thinking about it further this method is redundant. We have a default value property though, which is not being used anywhere though.

packages/datastore/datastore/src/ModelSchema.ts
wtrocki: Yep. Just remove password as example.

packages/datastore/datastore/src/ModelSchema.ts
wtrocki: It is critical for us to implement migrations (this will be required for CLI etc.) Keep it :D

PR #752, opened 2020-08-19
Datastore cli update

Eunovo:
### Description

WIP to generate and types and hooks.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

CC @ssd71 @wtrocki


Eunovo:
See https://github.com/aerogear/offix/blob/datastore-cli-update/packages/datastore/datastore-codegen/tests/mock.graphql
https://github.com/aerogear/offix/blob/datastore-cli-update/packages/datastore/datastore-codegen/src/generate.ts

ssd71:
Hi, I am looking at the code right now, and I can't seem to find anything wrong with it :(

The snapshots seem wrong though:
https://github.com/aerogear/offix/blob/cb0c5ae4e7b6b32f009587679f34ba7be24ffe41/packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap#L4-L10

Does the output schema show the exact same thing?

I will try to build and test this locally, but that will take some time because internet is crawling

In the meantime, CI seems to be failing on lint, if you could fix that and it still fails, I might be able to get something from there

Thanks

ssd71:
Thanks! The CI test is failing on snapshots, so this might be a problem with your locally installed packages, I will test locally and let you know regardless

ssd71:
I ran the tests locally and I have the exact same snapshots failing :confused:

Eunovo:
> Thanks! The CI test is failing on snapshots, so this might be a problem with your locally installed packages, I will test locally and let you know regardless

Yeah. I checked my version again. I'm running `yarn install` again

ssd71:
Also, `_version` is for conflicts, so if you want version in your model you have to enable conflicts for your models like so:
```typescript
new DataSyncPlugin({ modelConfigMap: { Comment: { enabled: true }, Note: { enabled: true } } })

**Comment by Eunovo** [comment]
Thanks @ssd71

**Comment by Eunovo** [comment]
@ssd71  What about `__deleted`? How do I get that field?

**Comment by ssd71** [comment]
The `_deleted` field is only available on the delta type and not on the base type

**Comment by wtrocki** [comment]
> The _deleted field is only available on the delta type and not on the base type

Yep. We use sync so we have access to deleted

**Comment by wtrocki** [comment]
> Also, _version is for conflicts, so if you want version in your model you have to enable conflicts for your models like so:

Ouch. We really need that global config and integration with graphql-serve. I will trigger new build so conflict will be possible with gql-serve

**Comment by Eunovo** [comment]
> Yep. We use sync so we have access to deleted

We only use the base model type currently, so we don't have `_deleted`

**Comment by wtrocki** [comment]
Ah.. right. Yep this might be bug in the server then. 
Why we need delta type when we can have _deleted on base type. Let's chat in the morning

File: packages/datastore/datastore-codegen/src/generate-documents/createJsonSchema.ts [hunk 1]
```diff
@@ -18,7 +18,7 @@ const getFieldParameters = (fieldName: string, type: GraphQLOutputType): any =>
         type = getNullableType(type);
         options.isRequired = true;
     }
-    if (fieldName === "id") {
+    if (type.toString() === "ID") {
```
wtrocki, line 21: The type will be GraphbackObjectID ?

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
```diff
@@ -0,0 +1,107 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`generate 1`] = `
+"export interface Note {
+    id: string;
+    title: string;
+    description: string;
+    comments: string[]
```
wtrocki, line 8: This probably makes no sense. We should detect relationships and decompose them. 
That is why Delta type is much better as it gives us that already.

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
```diff
@@ -0,0 +1,107 @@
...
+"
+`;
+
+exports[`generate 2`] = `
+"{
+  \\"Note\\": {
+    \\"name\\": \\"Note\\",
+    \\"version\\": 1,
+    \\"type\\": \\"object\\",
+    \\"primaryKey\\": \\"id\\",
```
wtrocki, line 36: This is always `_id`

Are you using latest graphback? 
Is that working for you?

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
```diff
@@ -0,0 +1,107 @@
...
+    \\"properties\\": {
+      \\"id\\": {
+        \\"type\\": \\"string\\",
+        \\"key\\": \\"id\\",
+        \\"isRequired\\": true,
+        \\"index\\": true,
```
wtrocki, line 42: The index on ID is redundant. The index is being used to apply extra indexes. ID do not need it.

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
```diff
@@ -0,0 +1,107 @@
...
+      },
+      \\"title\\": {
+        \\"type\\": \\"string\\",
+        \\"key\\": \\"title\\",
+        \\"isRequired\\": true
+      },
+      \\"description\\": {
+        \\"type\\": \\"string\\",
+        \\"key\\": \\"description\\",
+        \\"isRequired\\": true
```
wtrocki, line 53: Question. If we want to apply indexes. We might actually need some examples of overrides outside the generated source. If we override anything here it will be lost. 

I'm not sure what is the best approach here, but should be documented. 

Solution 1: copy this file outside and replicate the changes manually.
Solution 2: provide example for override in sample app.

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
```diff
@@ -0,0 +1,107 @@
...
+    }
+  }
+}"
+`;
+
+exports[`generate 3`] = `
+"import { ModelJsonSchema } from \\"offix-datastore\\";
+import jsonSchema from \\"./schema.json\\";
+
+type Schema<T = any> = {
```
wtrocki, line 99: Your source code is outdated. We have this in the core already.

Eunovo: I'll rebase

examples/react-datastore/src/datastore/generated/schema.json
```diff
@@ -3,6 +3,7 @@
     "name": "Todo",
     "version": 1,
     "type": "object",
+    "primaryKey": "_id",
```
wtrocki, line 6: This is not valid property. we setup this field in code from fields. Gianluca can confirm

wtrocki: Maybe we should have it on generator and remove it in code but this will allow people to have invalid values

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
wtrocki: Bumping this.

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
```diff
@@ -0,0 +1,101 @@
...
+}"
+`;
+
+exports[`generate 3`] = `
+"import { GeneratedModelSchema } from \\"offix-datastore\\";
+import jsonSchema from \\"./schema.json\\";
+
+export const schema = jsonSchema as GeneratedModelSchema;
+
+export * from \\"./types\\";
```
wtrocki, line 99: types?

wtrocki (changes requested):

docs/datastore/cli.md
```diff
@@ -4,56 +4,42 @@ title: Datastore CLI
 sidebar_label: Datastore CLI
 ---
 
-We provide a graphback plugin to generate required config for the DataStore.
+The Datastore CLI tool will generate JSON Schema files and types for your data models.
 
-## What is Graphback
+## Installing the CLI tool
 
-Graphback simplifies application development by generating a production-ready API
```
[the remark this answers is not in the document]
Eunovo: @wtrocki  Graphback intro was removed in later commits

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
Eunovo: Yes. This line(99) in generated/index.ts exports all types from generated/types

examples/react-datastore/src/datastore/hooks.ts
```diff
@@ -0,0 +1,11 @@
+import { Filter, useQuery, useSave, useUpdate, useRemove } from 'offix-datastore';
+import { TodoModel } from './config';
+import { Todo } from './generated';
+
+export const useFindTodos = (filter?: Filter<Todo>) => useQuery(TodoModel, filter);
```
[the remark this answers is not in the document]
Eunovo: We are not generating hooks currently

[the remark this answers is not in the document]
Eunovo: We talked about missing config problem on chat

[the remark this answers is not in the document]
wtrocki: I sent solutions

packages/datastore/datastore-codegen/tests/__snapshots__/Cli.test.ts.snap
wtrocki: This will not work IMHO. Schema is exported (but not going to get used anyway).
And types are precreated (but you cannot override schema).
We need to cover how to do overrides for that stuff. 

We need discussion about this. Maybe we should simply use GraphQL schema indexes and never allow to override schema.

examples/react-datastore/src/datastore/hooks.ts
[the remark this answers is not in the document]
wtrocki: Let's chat.

wtrocki:
Please let me know once this is ready.

docs/datastore/cli.md
```diff
@@ -4,56 +4,42 @@ title: Datastore CLI
...
-from data models to access data from one or more data sources.
-Graphback uses GraphQL and GraphQLCRUD to make it easy get the data you need,
-and follows the convention over configuration paradigm to to reduce the amount of
-setup and boilerplate costs associated with creating GraphQL applications.
+You can install `datastore-cli` globally with npm:
 
-Read about graphback-cli [here](https://graphback.dev/docs/cli/graphback-cli).
+`npm install -g datastore-cli`
```
wtrocki, line 13: Please rename to @offix/cli - datastore cli is too generic

PR #756, opened 2020-08-20
Native sqlite

kingsleyzissou:
### Description
- Native SQLite adapter
- Initial scaffolding for sample app

#### What's working:
- basic feature set
- tables are being created
- basic querying is working

#### What's not working:
- saving is breaking because the queue items are not being serialised before being saved to the queue

#### Not yet implemented:
- edit and delete


packages/datastore/datastore/src/storage/adapters/sqlite/SQLiteAdapter.ts
```diff
@@ -0,0 +1,221 @@
+import invariant from "tiny-invariant";
+// @ts-ignore
+import SQLite from "react-native-sqlite-2";
```
kingsleyzissou, line 3: The library doesn't have any type declarations, so I need to sort that out

wtrocki:
I think that after we merge other PR so many things will be fixed here. The challenge in this PR is that it brings sample app and other critical changes. I would love sample app to be pushed separately in PR it will be much easier to review as I see so many things that do not make sense to me and they probably come from sample app.

kingsleyzissou:
Closing as #783 has been merged

PR #787, opened 2020-08-26
Add capacitor config to sample app

kingsleyzissou:
### Description
I've added the basic config for the capacitor app, you'll just need to the path to your android studio in the config file


wtrocki:
Merging. We haven't tested SQLLite to full extent, but there are challenges we facing with the entire workflow that we need to address first on top of indexedDB provider.

PR #793, opened 2020-08-26
Swap client ids in queue

Eunovo:
### Description

This turned out to be much harder than expected. I'm still testing this with sample app and working out bugs.

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included

### DIFF
--- a/packages/datastore/datastore/src/replication/mutations/QueueUpdateProcessor.ts
+++ b/packages/datastore/datastore/src/replication/mutations/QueueUpdateProcessor.ts
@@ -1,0 +1,138 @@
+import { ModelMap } from "./MutationsQueue";
+import { MutationRequest } from "./MutationRequest";
+import { LocalStorage, CRUDEvents } from "../../storage";
+import { Model } from "../../Model";
+import traverse from "traverse";
+
+/**
+ * Swaps all client ids for server ids in queue
+ * and updates corresponding item in LocalStorage
+ */
+export class QueueUpdateProcessor {
+    private modelMap: ModelMap;
+    private storeName: string;
+    private clientData: any;
+    private serverData: any;
+    private storage: LocalStorage;
+    private currentModel: Model;
+
+    constructor(
+        modelMap: ModelMap,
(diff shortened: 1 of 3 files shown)


packages/datastore/datastore/src/replication/mutations/QueueUpdateProcessor.ts
```diff
@@ -0,0 +1,138 @@
...
+import { MutationRequest } from "./MutationRequest";
+import { LocalStorage, CRUDEvents } from "../../storage";
+import { Model } from "../../Model";
+import traverse from "traverse";
+
+/**
+ * Swaps all client ids for server ids in queue
+ * and updates corresponding item in LocalStorage
+ */
+export class QueueUpdateProcessor {
```
wtrocki, line 11: I think we will need to also handle addingNew item to queue from here to being able to add mutexes (separate PR/issue). Thx.

packages/datastore/datastore/src/replication/mutations/QueueUpdateProcessor.ts
```diff
@@ -0,0 +1,138 @@
+import { ModelMap } from "./MutationsQueue";
+import { MutationRequest } from "./MutationRequest";
+import { LocalStorage, CRUDEvents } from "../../storage";
+import { Model } from "../../Model";
+import traverse from "traverse";
```
wtrocki, line 5: why traverse is needed?

wtrocki (approved):

packages/datastore/datastore/src/replication/mutations/QueueUpdateProcessor.ts
Eunovo: It allows us to swap the client IDs in place anywhere in the queue items array

Eunovo:
@wtrocki this work still has bugs. I'm testing this in sample app

wtrocki:
No problem. I see them but want to work of the recent version when testing other storage as well

PR #863, opened 2020-09-28
fix: add ability to use target schema for model

wtrocki:
Remove version duplicates


kingsleyzissou:
@wtrocki I'll sort out the tests issue. Just need to regenerate the snapshots

kingsleyzissou (approved):

Issue #875, opened 2020-09-30
Using two the same hooks on the same page ends with render loop 


wtrocki:
## Bug Report

Steps to reproduce:

```
  const { error: errorTasks, data,
    subscribeToMore: subscribeToTasks } = useFindTasks(id);
  const { error: errorComments, data: dataComments,
    subscribeToMore: subscribeToComments } = useFindComments({ taskId: id });
  console.log(data);

```

After having both (even without subscription) page ends refreshing

wtrocki:
CC @kingsleyzissou

wtrocki:
Actually upon investigation this happens when there is wrong field passed to hook. Error handling does cause the loop.;

wtrocki:
Update problem happens when specifying filter (indexedDB) 

```
useFindComments({ taskId: id });
```

PR #877, opened 2020-10-01
fix: Loop on hooks

wtrocki:
This fixes loop on hook. Hook is called in loop because of data state changes (however strange thing happens when only id is present. This seems to solve issue. There might be more elegant solution though

#875


kingsleyzissou (approved):

wtrocki:
Actually reason why loop happens is return results. if single object is returned it doesn't rerender. Do we need to return results at all.. Probably not

PR #892, opened 2020-10-08
fix: make client arg optional in config

wtrocki:


kingsleyzissou (approved):

wtrocki:
@kingsleyzissou Can we do this in other PR. Tests were failing

Issue #872, opened 2020-09-30
Add relationships to schema generated by CLI.


wtrocki:
## Feature Request

Add relationships to models when generating them in the cli.

kingsleyzissou:
So just some clarification on this, if we have a model that looks something like this:

```
"""
@model
@datasync
"""
type Task {
  _id: GraphbackObjectID!
  title: String!
  description: String!
  status: TaskStatus
  type: String
  priority: Int
  public: Boolean
  startDate: GraphbackDateTime
  payload: GraphbackJSON

  """
  @oneToMany(field: 'note')
  """
  comments: [Comment]!
}
```

We want the schema to look somthing like this after using the cli:

```
"Comment": {
    ...
    "properties": {
       ...
       "noteId": {
        "type": "string",
        "key": "noteId",
        "isRequired": false,
        "relationship": "Task" // do we want to add this kind of field in the schema?
      },
    }
```

kingsleyzissou:
cc @wtrocki

wtrocki:
> `  "relationship": "Task"` 

- We need relationship (actual FK from db) to exist in the local database
- We need to be able to filter by this value (get all comments that are for this task)
- We need to be able to subscribe and get delta for this value (it should not change but yeah)
- We need to be able to swap ID for this relationship if the object was created using client-side ID.

We need this field in schema. We need it in types etc.

kingsleyzissou:
Cool thanks for clearing that up, that definitely makes sense. Just trying to get an idea of what the target schema will look like so we can work towards that

kingsleyzissou:
@wtrocki I actually think a made a bit of progress on this this evening. It's still a little bit tricky, but I think we can extract most of this information out of what is already there and add it to the final schema.

Do we want this to be bi-directional? So do we want a field on the task for comments with the relationship type and foreign key too? Or is it enough for it just to exist on the comment side for example?

kingsleyzissou:
This is still not working as expected, added to the MVP for the demo.

PR #899, opened 2020-10-12
WIP: generated relationship details for schema

kingsleyzissou:
### Description
This PR is a work in progress. The aim is to add relationship support to the generated schema as per #872

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included


kingsleyzissou:
@wtrocki it's a rough implementation for now, but it adds the field to the generated schema, as discussed in the issue.

For now it is only applying this in one direction, but it should be okay to modify this so we add details to both objects. It is also only applying these updates to the schema for now. I will add to this and will apply the changes to the generated types too.

packages/datastore/cli/src/generate-documents/createJsonSchema.ts
```diff
@@ -24,15 +21,33 @@ const getFieldParameters = (fieldName: string, type: GraphQLOutputType): any =>
 
 const getModelProperties = (model: ModelDefinition, primaryKey: string) => {
   const fieldMap = model.graphqlType.getFields();
+  const keys = Object.keys(fieldMap);
+
+  const relationships = model.relationships
```
wtrocki, line 26: This is amazing. We do not need to use delta type and getting all the information.

packages/datastore/cli/src/generate-documents/createJsonSchema.ts
```diff
@@ -24,15 +21,33 @@ const getFieldParameters = (fieldName: string, type: GraphQLOutputType): any =>
...
+    .filter(r => !keys.includes(r.ownerField.name))
+    .map(r => {
+      const fieldOptions = getFieldParameters(
+        r.relationForeignKey!,
+        r.relationType
+      );
+      fieldOptions.relationship = r.relationType;
```
wtrocki, line 33: @craicoverflow can you take look if this would work from your side. 
We actually getting model relationships and add extra fields to schema

wtrocki:
We need to fix listing and update snapshot

packages/datastore/cli/src/generate-documents/createJsonSchema.ts
craicoverflow: I'm not entirely sure yet, I think this might generate extra fields which are not part of a model. A `ModelDefinition` contains relationship information for the opposite relationship field.

For example, in a 1:M relationship between Note.comments and Comment.note, the `Note` model definition would have 2 relationships.

I think once we have a unit test in for this API, it will be easier to verify this.

kingsleyzissou: Yes you're correct @craicoverflow I had to filter out the side of the relationship that I didn't need here. But the relationship info went both ways oneToMany and manyToOne

craicoverflow: Is this the filtering you meant @kingsleyzissou ?

https://github.com/aerogear/offix/blob/schema-relationships/packages/datastore/cli/src/generate-documents/createJsonSchema.ts#L27

From looking at that I had thought it was filtering out the relationship information that Offix already had and did not want to apply twice.

kingsleyzissou: Yeah that's the filtering I meant @craicoverflow . As far as I can remember, we were having some issues with the comments array being included in our Datastore model and we needed to remove it.

We are only adding the foreign key information the the notes so we can filter those out based on the parent id, in this case the Task.

craicoverflow: See https://github.com/aerogear/offix/pull/900 - should this test be passing or am I doing something wrong?

kingsleyzissou: The test should be passing. I'm going to fix the linting and snapshot issues and then I can debug that

craicoverflow (commented):
From the side of Graphback relationship this looks good. I think having snapshots for each test case (just added along with the checking for defined bit) in `createJsonSchema.test.ts` would make viewing the output much easier though.

kingsleyzissou:
@craicoverflow good point, I can get that added. I need to update the snapshots now anyway because I had to remove the `_deleted` field (it was causing a 400 error).

wtrocki:
@kingsleyzissou do you mind doing release of the cli so I can test it with some complex schemas like OVP. Going to use cli as part of the demo :)

wtrocki:
After testing it I see that we do not filter by @datasync annotation. This means that we will query some id's from relationships:

```
      "distributionCentreId": {
        "type": "string",
        "key": "distributionCentreId",
        "relationship": "DistributionCentre"
      },
      "volunteerId": {
        "type": "string",
        "key": "volunteerId",
        "relationship": "Volunteer"
      },
```

but backend will not have them exposed as there is not @datasync on the type.

kingsleyzissou:
@wtrocki I've added that change we made before lunch. I will publish changes and will try verify with OVP.

wtrocki:
I think we need to build this into unit testing with snapshot here. 
Testing outside will not work in long term. Any issue we seen and case we handle should be in this repo

kingsleyzissou:
@wtrocki coming back to this now. It looks like we did some snapshots and unit tests for this. I think verifying outside was just to see if it worked. Are we okay to merge this?

wtrocki:
I need: 
1. Rebase.
2. npm publish --tag=dev
3. Then I will verify the changes in OVP and we can merge

wtrocki:
I would like us to use something like OVP schema as snapshot in near future

wtrocki (approved):

PR #900, opened 2020-10-13
test: add test to check if manyToOne relationship is added

craicoverflow:
I wanted to verify this so wrote a test which I think should pass but isn't. I don't have all the context so I could be wrong here. Take a look and see if this is expected or unexpected behaviour.

### DIFF
--- a/packages/datastore/cli/tests/createJsonSchema.test.ts
+++ b/packages/datastore/cli/tests/createJsonSchema.test.ts
@@ -1,0 +1,45 @@
+import { GraphbackCoreMetadata, ModelDefinition } from "@graphback/core";
+import { buildSchema } from "graphql";
+import { createJsonSchema } from "../src/generate-documents/createJsonSchema";
+
+test("createJsonSchema", () => {
+  const modelSDL =
+`
+scalar GraphbackObjectID
+
+"""
+@model
+@datasync
+"""
+type Note {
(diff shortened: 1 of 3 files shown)


packages/datastore/cli/tests/createJsonSchema.test.ts
```diff
@@ -0,0 +1,45 @@
...
+"""
+@model
+@datasync
+"""
+type Note {
+  _id: GraphbackObjectID!
+  title: String!
+  description: String!
+  """
+  @oneToMany(field: 'noteComment', key: 'noteId')
```
wtrocki, line 19: I will also add one with `noteID` missing

kingsleyzissou:
@craicoverflow Thanks for this. Let me fix the lint issues on my end first because I can't see the failing tests.

kingsleyzissou:
@craicoverflow the issue is with the filter method I wrote. It was using the default Type name for filtering out the undesired properties. When you specified the fieldname here: `@oneToMany(field: 'noteComment', key: 'noteId')`, it broke things.

What I will do, is I will merge this into my branch and refactor, because this test _should_ pass.

kingsleyzissou (approved):

craicoverflow:
Sure - request my review whenever you need me to take another look.

PR #982, opened 2020-12-07
Apply late binding for model replication (fixes #891)

### LINKED ISSUE #891: Ability to provide late binding for datastore queries.
## Feature Request

The query for datastore might not be known at the moment of creating the store. 
For example, logged user ID needs to be retrieved etc.

kingsleyzissou:
### Description
We need to provide the ability to start the replication at a later point in the app start up rather than only during initialisation.
At present, if we need to filter by a user id for example, there is no way to do this at a network level and we can only filter the results from the datastore. This will be an issue in a larger scale app.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/packages/datastore/datastore/src/ModelSchema.ts
+++ b/packages/datastore/datastore/src/ModelSchema.ts
@@ -49,6 +49,11 @@ export declare class ModelJsonSchema<T> {
    */
   namespace?: string;
   /**
+   * Flag for whether or not model replication
+   * should be initialised at a later point
+   */
+  lateInit?: boolean;
+  /**
    * Model version number
    */
   version?: number;
@@ -89,6 +94,7 @@ export declare class ModelJsonSchema<T> {
 export class ModelSchema<T = any>{
   private name: string;
   private namespace: string;
+  private lateInit: boolean;
   private primaryKey: string;
   private fields: Fields<T>;
   private encrypted: string[];
@@ -100,6 +106,7 @@ export class ModelSchema<T = any>{
     this.version = schema.version || 0;
     this.name = schema.name;
     this.namespace = schema.namespace || "user";
+    this.lateInit = schema.lateInit || false;
     this.fields = extractFields(schema);
     this.primaryKey = extractPrimary(this.fields, schema.primaryKey);
     this.indexes = extractIndexes(this.fields, schema.indexes);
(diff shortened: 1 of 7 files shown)


kingsleyzissou:
@wtrocki I think we are going to need the ability to stop replication and restart. Say for example someone logs out and signs in with a different user account, we need to be able to update the filters and restart the replication. We can't do that with the current architecture at all

wtrocki:
@kingsleyzissou Exactly. +1 for current architcture. but we might actually need some interface that represents some actions we can control.

kingsleyzissou:
@wtrocki so I had to add a stop method to the subscription replicator. At the moment the update filter is on the model, but this can be changed quite easily. I have pushed some changes, I want to publish a new release so I can test on OVP.

wtrocki:
Let's publish. Looks good but would like to see this in action

kingsleyzissou:
@wtrocki published 0.3.7 and tested it with OVP. It doesn't look like the filters are updating properly. Will have to do some debugging.

wtrocki:
What branch on OVP we are using

kingsleyzissou:
@wtrocki sorry I missed your message, I'm using the demooffix branch

kingsleyzissou:
@wtrocki This is working now, I am busy creating a PR in OVP so you can verify. I'll add verification steps here and
I will update docs too.

kingsleyzissou:
@wtrocki, can't do the docs until I've updated the CLI. For now I'm editing the generated schema and adding the property `lateInit: true`, which means that the model initialisation will be deferred until that is manually started

packages/datastore/datastore/src/ModelSchema.ts
```diff
@@ -48,6 +48,11 @@ export declare class ModelJsonSchema<T> {
    * default is `user_<Model name>`
    */
   namespace?: string;
+  /**
+   * Flag for whether or not model replication
+   * should be initialised at a later point
+   */
+  lateInit?: boolean;
```
wtrocki, line 55: Flag for behaviour is not desired. Any reason we do not always want late init

wtrocki:
Not a big fan of late init as this goes against simplification target. Would be possible to do lateInit=true and remove flag? any side effects?

kingsleyzissou:
@wtrocki I think I was having issues with doing `datastore.init()` at a later stage. It was okay for the fist time you changed things, but if you stopped and started again it would hang. Maybe what I could do is decouple things a bit and have an init method for the replicator and a start replication method which could be invoked at a later stage. Wdyt?

wtrocki:
I have played with this today and it seems like:
Init should initialize datatabase and perform migrations.
Replication should be triggered manually and you should specify filter explicitly (along with the config). Specifying other filter should replace stop (keep things simple). If we want to stop replication for specific model that could be overly complicating things so I would suggest global stop instead (like stop on logout)

PR #1044, opened 2021-02-05
fix(conflictHandler): type check on object attribute to allow null value

mbrechet:
The conflict handler failed to detect conflict when null attribute are present

### Description

The correction test `typeof(base[key])` object's attribute to check the existence of the attribute on object instead of just check `base[key]`

##### Checklist

- [x] `npm test` passes
- [x] `npm run build` works
- [x] tests are included


kingsleyzissou (approved):
@mbrechet thanks for this!

mbrechet:
@kingsleyzissou  you're welcome ! :smiley:

offix is amazing :+1: 

Is it possible to publish a pre-release beta to get the fix from npm ? 

Thanks

PR #1070, opened 2021-02-25
Datastore network indicator (fixes #1069)

### LINKED ISSUE #1069: Expose network status change
## Feature Request

I would like to expose the network status indicator used in the datastore, I just wanted to start the discussion first. It would be nice to hook into the implemented datastore so we can subscribe to network status changes on the client side too without having to create a new implementation of the network indicator.

kingsleyzissou:
### Description
- Provide the ability to subscribe to datastore network status changes.

##### Checklist

- [ ] `npm test` passes
- [ ] `npm run build` works
- [ ] tests are included

### DIFF
--- a/examples/react-datastore/package.json
+++ b/examples/react-datastore/package.json
@@ -16,7 +16,7 @@
     "@types/react-dom": "^16.9.0",
     "antd": "~4.4.1",
     "graphql.macro": "^1.4.2",
-    "offix-datastore": "0.3.17",
+    "offix-datastore": "0.3.19",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-scripts": "3.4.1",
--- a/examples/react-datastore/package.json
+++ b/examples/react-datastore/package.json
@@ -16,7 +16,7 @@
     "@types/react-dom": "^16.9.0",
     "antd": "~4.4.1",
     "graphql.macro": "^1.4.2",
-    "offix-datastore": "0.3.19",
+    "offix-datastore": "0.4.0",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-scripts": "3.4.1",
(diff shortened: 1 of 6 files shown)


kingsleyzissou:
Published `0.3.19`

examples/react-datastore/package.json
```diff
@@ -16,7 +16,7 @@
     "@types/react-dom": "^16.9.0",
     "antd": "~4.4.1",
     "graphql.macro": "^1.4.2",
-    "offix-datastore": "0.3.17",
+    "offix-datastore": "0.3.19",
```
wtrocki, line 19: Let's do 0.4.0 :D

examples/react-datastore/src/App.tsx
```diff
@@ -5,30 +5,43 @@ import 'antd/dist/antd.css';
...
+    datastore.startReplication()
+  });
 
-  const toggleReplication = () => {
-    if (replicating) {
-      datastore.stopReplication()
-    } 
-    setReplicating(!replicating)
-  }
+  // const toggleReplication = () => {
```
wtrocki, line 39: Why this is disabled

kingsleyzissou: Oh should have deleted. I removed it now because I have the change in network status stopping and starting the replication automatically

examples/react-datastore/package.json
kingsleyzissou: Awesome will do that now

examples/react-datastore/src/App.tsx
wtrocki: That should happen anyway in the core right?

wtrocki (approved):

examples/react-datastore/src/App.tsx
kingsleyzissou: Yeah, this is more for the OVP stuff, because turning switching to offline mode in chrome doesn't really stop the replication or subscriptions from coming in.

